PR #13, opened 2026-08-29
feat(scorers): add evidence-backed groundedness

### LINKED ISSUE #12: Add a groundedness scorer for RAG outputs
First slice of #4: a scorer in `rai_toolkit/scorers/` that grades whether each response is supported by the retrieved context supplied with the dataset row. Rows without retrieved context should be marked un-assessed rather than defaulted (same convention as `RubricScorer`). Output should carry the supporting/contradicting spans as evidence so reports stay evidence-backed. Follow-ups tracked separately: retrieval relevance and context precision/recall.

adity982:
Closes #12

## Summary

- add a `GroundednessScorer` that grades responses only when retrieved context is present
- mark missing-context rows explicitly un-assessed, matching the existing rubric convention
- retain only verbatim supporting and contradicting spans that can be verified against the row
- add mocked offline tests for supported, contradicted, fabricated-evidence, and missing-context paths

## Validation

- `pytest tests/test_groundedness_scorer.py -q` (3 passed)
- `ruff check --ignore BLE001` on changed Python files
- `ruff format --check` on the new test, prompt registry, and scorer exports
- `git diff --check` on task files

The checkout also contains an unstaged CRLF-only change in `rai_toolkit/redteam/attacks.py`; it was excluded from the commit and PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,10 @@
 !/integrations/
 !/integrations/**
 
+# Test sources are published with the repository but excluded from wheels.
+!/tests/
+!/tests/**
+
 # Reference reviewer UI and example apps (see README).
 !/demo/
 !/demo/**
--- a/README.md
+++ b/README.md
@@ -369,6 +369,12 @@ LLM-as-judge scorers load prompt templates from `rai_toolkit/prompts/`
 (`judge_prompts.py`). Extend or override prompts there when tuning judges for
 a domain without forking scorer code.
 
+Available judge scorers include `FactualityJudge`, `FairnessJudge`,
+`ContentSafetyJudge`, `PrivacyJudge`, `SecurityJudge`, `TransparencyJudge`,
+`ExplainabilityJudge`, `RubricScorer`, and `GroundednessScorer`. The
+`GroundednessScorer` evaluates RAG responses only when retrieved context is
+present and retains verified supporting or contradicting spans as evidence.
+
 ## Weave-native evaluation in assessment
 
 When you pass `weave_project=` to `Assessor` (or `rai assess --weave-project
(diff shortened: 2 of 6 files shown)


knisar:
Thanks @adity982 , this is exactly what #12 was after. Good calls on marking missing-context rows un-assessed to match the rubric convention, and on keeping only verbatim spans that verify against the row. The offline mocked tests are appreciated. Two things for when you mark it ready: add the scorer to the README scorer list, and confirm it's exported so the presets can reference it. Ping me and I'll do a full review pass.

adity982:
@knisar Addressed in `06eb42a`: the README scorer list now includes `GroundednessScorer` and documents its context/evidence behavior. I also confirmed it is publicly imported and included in `rai_toolkit.scorers.__all__`; the existing focused test imports it from `rai_toolkit.scorers`, so presets can resolve that public export.

Fresh validation: `pytest tests/test_groundedness_scorer.py -q` (3 passed), direct public-export assertion passed, README contract check passed, and `git diff --check -- README.md` passed. I am marking the PR ready for your full review pass.

knisar (approved):
Reviewed the diff against the scorer base classes. The un-assessed path matches the FactualityJudge/RubricScorer convention (assessed=False, excluded from gates and aggregations), from_compliance_scale clamps the raw score so an out-of-range judge response can't break ScorerResult, and the span verification fails in the safe direction: anything not verbatim from the row is discarded and surfaces in discarded_evidence_spans. Tests cover the three paths that matter, and the README and export wiring are in.

Two follow-ups I'll open as separate issues rather than block this PR: skipping behavioral-refusal rows the way FactualityJudge does, and exact-substring matching over-discarding legitimate spans when the judge normalizes whitespace or quotes.

Thanks @adity982, great first contribution. Merging.

PR #17, opened 2026-08-30
fix: skip behavioral refusals in GroundednessScorer

### LINKED ISSUE #15: GroundednessScorer should skip behavioral-refusal rows like FactualityJudge
FactualityJudge short-circuits rows whose `expected` field indicates a refusal/boundary probe (via `_is_behavioral_refusal_expected`) so a correct refusal is not scored as a grounding failure. `GroundednessScorer` (#13) only short-circuits on empty context, so a refusal response on a row that carries retrieved context is graded as ungrounded and fails the row.

Fix: apply the same guard at the top of `GroundednessScorer.score()` before the empty-context check, returning the standard un-assessed result (`details["skipped"] = "behavioral_refusal_expected"`).

Test: a refusal-expected row with non-empty context returns `assessed=False` and never calls the judge.

a0927929980-dot:
## Summary

- skip behavioral-refusal rows before the empty-context check
- return the standard un-assessed groundedness result
- add a regression test that verifies the judge is not called

Fixes #15

## Testing

- `python -m pytest -q` — 4 passed
- `python -m compileall -q rai_toolkit tests`
- `python -m pip check`
- `reuse lint`

### DIFF
--- a/rai_toolkit/scorers/llm_judges.py
+++ b/rai_toolkit/scorers/llm_judges.py
@@ -317,6 +317,27 @@ class GroundednessScorer(LLMJudgeScorer):
         context: str = "",
         **kwargs: Any,
     ) -> ScorerResult:
+        expected = str(kwargs.get("expected") or "")
+        if _is_behavioral_refusal_expected(expected):
+            return ScorerResult(
+                score=0.0,
+                passed=False,
+                category=self.category,
+                explanation=(
+                    "Un-assessed: this row expects refusal or boundary-setting "
+                    "behavior, not grounded factual claims. GroundednessScorer "
+                    "does not penalize correct safety refusals; use the relevant "
+                    "privacy/security/safety scorer instead."
+                ),
+                details={
+                    "skipped": "behavioral_refusal_expected",
+                    "scorer_name": self.name,
+                    "judge_model": self.model,
+                    "supporting_spans": [],
+                    "contradicting_spans": [],
+                },
+                assessed=False,
+            )
         if not context.strip():
             return ScorerResult(
                 score=0.0,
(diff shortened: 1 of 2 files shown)


knisar (changes requested):
Thanks @a0927929980-dot, the guard matches #15 and mirrors the FactualityJudge pattern correctly, including the span keys in details.

One addition before I merge: the refusal markers in _is_behavioral_refusal_expected are substring-based, so I want a negative test pinning that the guard does not over-trigger. Please add a test where a row has context AND an ordinary factual expected value (for example "Answer with the revenue figure from the context") and assert the judge IS called and the result is assessed. That protects normal rows that happen to carry an expected field.

Out of curiosity, how did you come across the project, and are you using the toolkit somewhere? Always interested in hearing what contributors are building with it.

a0927929980-dot:
Thanks for the review and for pointing this out. That makes sense. I’ll add a negative regression test with non-empty context and an ordinary factual `expected` value, and verify that the judge is called and the result is assessed. I’ll update the PR once it’s ready.

I found the project with help from Codex while looking for active, newcomer-friendly open-source issues so I could learn the full contribution workflow. I’m not using the toolkit in production yet; I’m currently exploring it to learn more about responsible AI evaluation and guardrail testing. Thanks for the opportunity to contribute!

a0927929980-dot:
I've added the requested negative regression test in commit 99bd480. It uses non-empty context and an ordinary factual expected value, and verifies that the judge is called and the result is assessed. All 5 tests pass. Thanks!

knisar (approved):
The negative test was the missing piece, thanks. Guard placement and the details shape both match the FactualityJudge convention, and the new test pins exactly the over-trigger case I was worried about. Merging. Thanks also for the honest answer on how you found the project, and good luck with the rest of the learning, this was done properly.

a0927929980-dot:
Thank you for the kind words and for merging the PR! I really appreciate your review and guidance. Adding the negative test helped me better understand how to protect normal factual rows from false refusal matches. I learned a lot from this contribution and hope to contribute to the project again in the future.

PR #18, opened 2026-08-30
fix: accept normalized groundedness evidence spans

### LINKED ISSUE #16: Evidence spans are over-discarded when the judge normalizes whitespace or quotes
`_verified_evidence_spans` (added in #13) requires exact substring membership in the row's output/context. Judge models often collapse whitespace, trim newlines, or emit curly quotes, so legitimate spans get discarded and inflate `discarded_evidence_spans`.

Discarding is the right failure direction, so keep that default. But a light normalization pass (collapse internal whitespace, normalize quote characters) applied to both the span and the row text before the membership check would cut false discards without weakening the verbatim guarantee.

Constraint: the spans stored in `details` must remain verbatim text from the row, not the judge's version. Locate the match in the normalized space, then slice the original row text for storage.

nightcityblade:
Fixes #16

## Summary

- Normalize internal whitespace and typographic quotes before verifying judge evidence spans.
- Map normalized matches back to original row offsets so stored evidence remains verbatim.

## Tests

- `.venv/bin/python -m pytest -q` — 4 passed
- `ruff check rai_toolkit/scorers/llm_judges.py tests/test_groundedness_scorer.py` — passed

### DIFF
--- a/rai_toolkit/scorers/llm_judges.py
+++ b/rai_toolkit/scorers/llm_judges.py
@@ -290,7 +290,11 @@ def _normalized_text_with_offsets(text: str) -> tuple[str, list[int]]:
 
 
 def _verbatim_span(raw_span: str, row_text: str) -> str | None:
-    normalized_span, _ = _normalized_text_with_offsets(raw_span.strip())
+    raw_span = raw_span.strip()
+    if raw_span and raw_span in row_text:
+        return raw_span
+
+    normalized_span, _ = _normalized_text_with_offsets(raw_span)
     normalized_row, offsets = _normalized_text_with_offsets(row_text)
     start = normalized_row.find(normalized_span) if normalized_span else -1
     if start < 0:
--- a/rai_toolkit/scorers/llm_judges.py
+++ b/rai_toolkit/scorers/llm_judges.py
@@ -270,6 +270,34 @@ class FactualityJudge(LLMJudgeScorer):
         return super().score(output=output, input=input, context=context, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


knisar (changes requested):
Thanks @nightcityblade, and thanks for claiming the issue on the thread before opening this. The offset-map approach is right, the verbatim-row-text constraint is honored, and the 8-character quote table is more complete than I expected.

Two changes before I merge:

1. Test through the public path instead of importing the private helper. Build a GroundednessScorer with a mocked _call_judge (see the _scorer helper already in the test file) and assert on result.details, the way the neighboring tests do. Tests reaching into _verified_evidence_spans will break on any internal refactor.

2. Add a cheap fast path in _verbatim_span: if the raw span is already a substring of the row text, return it before doing the normalization work. Exact matches are the common case.

Also note #19 landed on this same issue minutes after yours; since you claimed it first on the thread, this PR gets the review. Ping me when the changes are in and I'll do the merge pass.

nightcityblade:
Addressed both requested changes in `6200354`:

- the normalized-span regression now exercises `GroundednessScorer.score()` and asserts on `result.details`, without importing the private helper
- `_verbatim_span()` returns exact raw substrings before building normalized text and offset maps

Validation: `pytest tests/test_groundedness_scorer.py -q` (4 passed), Ruff check on both changed files, and `git diff --check` all pass. Ready for the merge pass when convenient—thank you.

knisar:
Both changes look right. The regression going through score() means it survives any internal refactor, and the fast path covers the common case cheaply. The offset-map approach held up on second read. Merging, and thanks again for claiming on the thread before opening the PR.

PR #19, opened 2026-08-30
fix(scorers): match evidence spans through judge-side normalization

### LINKED ISSUE #16: Evidence spans are over-discarded when the judge normalizes whitespace or quotes

cyclomati:
Closes #16

`_verified_evidence_spans` required exact substring membership, so spans the judge lightly normalized — collapsed whitespace, straightened quotes — were discarded and inflated `discarded_evidence_spans`.

**Approach** (as proposed in the issue):

- `_normalized_chars` collapses whitespace runs to a single space and unifies curly/straight quote characters, returning the normalized text **plus an index map back to the original**.
- `_find_verbatim_span` keeps the exact-substring fast path; on miss it searches in normalized space and slices the match **from the original row text**, so stored evidence remains verbatim row text, never the judge's version.
- Discarding stays the failure direction: spans that don't match even after normalization are still dropped, and fabricated spans are unaffected.

**Tests:** two new cases — a context span with a collapsed newline+indentation run, and straight-vs-curly quotes — both asserting the *stored* span is the original row text and `discarded_evidence_spans == 0`. Existing tests unchanged and passing (5/5).

### DIFF
--- a/rai_toolkit/scorers/llm_judges.py
+++ b/rai_toolkit/scorers/llm_judges.py
@@ -270,6 +270,61 @@ class FactualityJudge(LLMJudgeScorer):
         return super().score(output=output, input=input, context=context, **kwargs)
 
 
+_QUOTE_EQUIVALENTS = {
+    "‘": "'",
+    "’": "'",
+    "“": '"',
+    "”": '"',
+}
+
+
+def _normalized_chars(text: str) -> tuple[str, list[int]]:
+    """Normalize for matching: collapse whitespace runs, unify quote characters.
+
+    Returns the normalized text together with a map from each normalized
+    character back to its index in the original text, so a match found in
+    normalized space can be sliced verbatim from the original.
+    """
+
+    chars: list[str] = []
+    positions: list[int] = []
+    space_pending = False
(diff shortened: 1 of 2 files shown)


knisar:
Thanks for this, it's a good implementation. Unlucky timing though: nightcityblade claimed #16 on the issue thread before either PR went up, and #18 covers the same fix, so that one gets the review and I'm closing this as a duplicate.

If you're up for more, comment on an issue to claim it before you start. That's how I'll resolve collisions like this one going forward. More scorer and framework-mapping issues are coming.

PR #25, opened 2026-09-01
feat: add RetrievalRelevanceScorer for RAG retrieval quality evaluation

### LINKED ISSUE #4: Add RAG and agent-specific evaluation metrics
It would be valuable to include first-class evaluation support for RAG and agentic systems. Useful areas could include retrieval relevance, context precision/recall, groundedness, citation correctness, tool-call accuracy, task completion, trajectory quality, and agent policy compliance.

Zeming-Yuan:
Closes #4 (retrieval relevance track)

Adds `RetrievalRelevanceScorer`, an LLM-as-a-Judge scorer that evaluates
whether each retrieved context chunk is relevant to the user query. Designed
for RAG retrieval quality evaluation alongside the existing `GroundednessScorer`.

**Design:**
- Context string contains multiple chunks separated by `---`
- Judge assigns per-chunk relevance verdicts (`relevant`, `partially_relevant`, `irrelevant`)
- Overall score on the 0-3 compliance scale, normalized to 0-1
- Follows established conventions: behavioral-refusal rows return `assessed=False`,
  empty context returns `assessed=False`, tests mock `_call_judge`

**Changes:**
- `rai_toolkit/prompts/judge_prompts.py`: prompt template for retrieval relevance
- `rai_toolkit/scorers/llm_judges.py`: `RetrievalRelevanceScorer` class
- `rai_toolkit/scorers/__init__.py`: export in import and `__all__`
- `tests/test_retrieval_relevance.py`: 7 tests covering all code paths

All 13 tests pass (6 existing + 7 new).

Developed with AI assistance.


knisar:
Before this goes any further: the diff on this PR is +287 / -17,477 across 107 files. It deletes demo/, demo_app/, integrations/, docs/, LICENSES/ and most of rai_toolkit/.

Your actual scorer work looks like the 287 additions. I think the branch was cut from an early commit rather than current main, so everything added to main since then is showing up as a deletion.

From your side:

```
git remote add upstream https://github.com/wandb/rai-toolkit.git
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

The PR updates in place, and it should then show only the prompt template, the scorer class, the export, and the tests.

Once the diff is clean I will review the substance. The plan you posted on #4 is fine, this is purely a branch problem.

Zeming-Yuan:
@knisar Done — rebased onto current main. The diff is now +287 / 0 across 4 files only:

- `rai_toolkit/prompts/judge_prompts.py` — judge prompt template
- `rai_toolkit/scorers/llm_judges.py` — RetrievalRelevanceScorer class
- `rai_toolkit/scorers/__init__.py` — export
- `tests/test_retrieval_relevance.py` — 8 test cases

Ready for review.

knisar:
Rebase worked. Thanks for turning that around so fast. +287 / 0 across four files is the shape I was hoping for, and you hit all three registration points, so nothing is orphaned.

Conventions look right too: refusal check ahead of the empty-context check, both returning `assessed=False` with a `skipped` reason, `MIT-3.1` to line up with `GroundednessScorer`. I also like that the system prompt rules out topically adjacent chunks explicitly. That is usually where relevance judges go soft.

All 7 tests pass here, and so does the full suite. Two things I want fixed before this goes in.

**1. The JSON example in your template has four braces where it needs two.**

`_format_prompt` calls `.format()` once, so `{{` collapses to `{`. That is why `GROUNDEDNESS_SCORER_TEMPLATE` writes `{{"score": <0-3>, ...}}`. Yours renders as `{{ "score": ... }}`, and that is what the judge is being told to imitate:

```bash
python -c "from rai_toolkit.prompts.judge_prompts import JUDGE_PROMPTS; print(JUDGE_PROMPTS['RetrievalRelevanceScorer']['template'].format(output='o', input='i', context='c'))"
```

Two braces everywhere, including the nested verdict object. The tests do format the template, they just never look at what comes out, which is why nothing caught it. A test that formats it and asserts the JSON block parses would be a good thing to leave behind for whoever writes the next scorer.

**2. Chunk verdicts land in the report unchecked, and the empty case passes.**

`total_chunks` and `relevant_chunks` are both counted off the judge's list rather than off the context, which the scorer never splits on `---`. Two consequences:

- Three chunks in, a judge that returns five verdicts including `chunk_index: 7`, and details reports `relevant_chunks: 5, total_chunks: 5`. The score itself survives, since that comes from the judge's separate `score` field, but details and anything derived from them are wrong.
- No verdicts at all gives `assessed=True, score=1.0, passed=True, relevant_chunks=0, total_chunks=0`. A silent judge reading as perfect retrieval is the one result this scorer cannot be allowed to produce.

`GroundednessScorer` deals with the same problem by running judge-claimed spans through `_verified_evidence_spans` and reporting `discarded_evidence_spans`. Same idea here: split the real context on `---` and take `total_chunks` from that rather than from the judge, keep only verdicts that are dicts with a unique in-range integer index and a label you recognize, report what you dropped and what is missing, and return `assessed=False` with `skipped="judge_parse_failure"` if any real chunk did not get one.

Heads up that this means editing two existing tests, not only adding new ones. `test_no_relevant_chunks_scores_zero` asserts `total_chunks == 1` on a two-chunk context, and `test_empty_chunk_verdicts_handled` expects an assessed result with nothing back from the judge. Both bake in the behavior above. Duplicate indices, out-of-range indices, unknown labels and partial coverage would all be worth a case.

Smaller things, and the last two are your call. `float(result.get("score", 0))` turns `"NaN"` into 1.0 and passes it, while `"high"` raises, so a finite 0 to 3 check returning an unassessed parse failure would close both. Context that is only a delimiter, and blank queries, currently get assessed rather than skipped. And I would drop `details["judge_response"]` on the success path once you have validated verdicts to report; Groundedness leaves the raw payload out, and these dicts get written into report files.

One note on the prompt itself: it does not say how `partially_relevant` feeds the overall score, or what an exactly half-relevant retrieval should get. A line either way would help the judge stay consistent.

Last thing, and it is not yours. If you saw `scorer.category` come back empty when constructing the scorer directly, that was #28. #30 fixed it and it is on main now, so rebase and your `MIT-3.1` will stick.

None of this is structural, it is all inside your four files. Push when you have it and I will look the same day.

Zeming-Yuan:
Thanks for the thorough review — all fixed and pushed.

**Template braces**: the JSON example now uses single-brace escaping, and your one-liner renders `{"score": ...}` cleanly. Left behind `test_template_json_example_is_parseable`, which formats the template and json-parses the example block (placeholders filled the way a judge would) so the next scorer author inherits the check.

**Verdict validation**: `total_chunks` now comes from splitting the real context on `---`. Verdicts are kept only if they are dicts with a unique in-range integer `chunk_index` and a recognized label; duplicates keep the first verdict, everything else lands in `details["discarded_verdicts"]`. If any real chunk lacks a verdict the row returns `assessed=False` with `skipped="judge_parse_failure"`, carrying `covered_chunks`, `missing_chunk_indexes` and the raw `judge_response`.

**Score check**: `raw_score` now must be a finite 0-3 number — "NaN", "Infinity", 5, -1, "high" and null all return the unassessed parse failure (parametrized in one test).

**Smaller items**: delimiter-only context and blank queries skip as unassessed (`empty_context` / `empty_query`); `details["judge_response"]` is dropped on the success path; the prompt now states that `partially_relevant` counts as half a relevant chunk when weighing the overall score, so an exactly-half retrieval scores 1.

One heads-up: rebasing onto main for #30 meant the validation also caught a third existing test — `test_factual_expected_is_assessed_and_calls_judge` mocked a judge response with no verdicts, so it now carries a valid verdict set to stay assessed. The two tests you flagged are updated as described (`total_chunks == 2` on the two-chunk context; the empty-verdicts case now expects the parse failure).

Full suite green here: 35 passed.

PR #27, opened 2026-09-01
feat(scorers): add citation correctness scorer for RAG outputs

### LINKED ISSUE #22: Add a citation correctness scorer for RAG outputs
Split from #4 and being implemented in #27 by @M4h1m4. Follows the shape of #12.

A CitationCorrectnessScorer in rai_toolkit/scorers/ that checks whether citations in a response resolve to retrieved context that actually supports the cited claim. This catches two failures GroundednessScorer does not: a citation pointing at a source absent from the context, and a supported claim attributed to the wrong source. A response can be fully grounded and still cite incorrectly.

Conventions agreed on the #4 thread:

- Cited spans are emitted verbatim and verified against the row before being trusted, reusing _verified_evidence_spans as-is (post-#18 it handles whitespace/quote normalization with verbatim mapping back to row offsets).
- Keep the existing {response_span, context_span} span shape and carry the citation marker alongside it.
- Behavioral-refusal rows are skipped ahead of the empty-context branch, matching the FactualityJudge ordering and #15.
- Rows with no retrieved context return assessed=False, not a defaulted zero.
- A response with no citations returns un-assessed with skipped: "no_citations". Citation presence enforcement is a separate policy metric, out of scope here.
- Tests mock _call_judge, no key or network needed.

Retrieval relevance and context precision/recall stay on #4.

M4h1m4:
Closes #22. Part of #4.

Adds `CitationCorrectnessScorer`, which checks whether each cited claim is supported by the source it names.

## Why this is not `GroundednessScorer`

Groundedness asks *"is this claim supported by the context at all?"* This asks *"does the source the claim points at support it?"* Two failures groundedness passes:

| Failure | Example |
|---|---|
| **Fabricated citation** | Response cites `[reg-z-2024]`; no such source is in the retrieved context |
| **Misattribution** | The claim is true and *is* in the context, but credited to the wrong block |

A response can be fully grounded and still cite incorrectly. `MIT-3.1` names this risk directly: *"inaccurate, fabricated, or misleading information including hallucinations, **false citations**, invented facts, and unsupported claims."*

## How it works

The unit of judgment is the **citation occurrence** — one claim, citing one source, at one position — not the marker. Two claims citing the same source are two things to verify.

Deterministic work happens in Python; the judge is asked only the question that needs judgment.

| Task | Who | How |
|---|---|---|
| Where are the citations? | Python | regex over the response |
| Which name a real source? | Python | parsed `[source-id]` blocks in the context |
| Does the named block support the claim? | **Judge** | one verdict per occurrence |
| Is the judge's evidence real? | Python | span must be verbatim in the right block |
| Does the score match its own verdicts? | Python | band check |

### Classification

Context is split into `[source-id] text` blocks. Each citation occurrence falls into one of four buckets:

- **resolved** — names a parsed block; goes to the judge
- **ambiguous** — appears bracketed in the context but not as a usable block label; the parser is the likely culprit, so no accusation is made
- **fabricated** — appears nowhere in the context **and** matches a label style this context uses distinguishably
- **ignored** — does not look like a source id at all (`arr[0]`, `[TODO]`, markdown-link text, anything inside code)

The shape comparison uses a signature of `(is_numeric, has_hyphen, has_dot, has_underscore)` and is decided **per label style**, not for the context as a whole. Where a style carries no structure — bare words, bare numbers — nothing of that shape is accused, because `[x]` and `arr[0]` are indistinguishable from such labels. A signature match alone is not enough: `[0-1]` and `[x-y]` match `[doc-1]` exactly, so a marker must also carry a part that reads as a word.

### What the judge sees

The response is shown with an occurrence tag after each gradeable citation:

```
The rule [adverse-action]⟦1⟧ requires notices, and deadlines apply [fair-lending]⟦2⟧.
```

The judge reads the prose as written and returns one verdict per number. This scorer never decides where a claim begins or ends — deriving that boundary truncated mid-sentence citations to a fragment and emptied ones that opened a sentence. Binding stays exact because the number is attached to the marker in the text.

The delimiter pair is **chosen per response** from a candidate list, taking the first pair absent from it, so annotation is purely additive: removing the inserted tags returns the response verbatim. A fixed pair had to be stripped on collision, and stripping it kept what sat between, so `the rate is ⟦1⟧5%` reached the judge as `15%`. The prompt names the pair actually used, since it is no longer fixed.

`response_span` is advisory: recorded when it quotes the response verbatim, because a report reads better naming the claim than an index, but it never binds a verdict. A swapped quote cannot move a verdict between citations, and an unverifiable one is omitted while the verdict stands on its verified context evidence.

### Before any API call

Six guards short-circuit before the judge is called: behavioural-refusal rows, empty context, a response with no citations, a label style that cannot support an accusation, citations that resolve to nothing, and partial coverage. The refusal check precedes the empty-context check, matching `FactualityJudge` and the ordering fixed in #15.

### After the reply

Nothing becomes an assessed score until the reply is usable, complete and self-consistent:

- the score must be a finite number in 0–3, and not a boolean
- exactly one verified outcome per occurrence; a second is a contradiction, not something to resolve by taking whichever verified
- every resolved occurrence must carry a verdict whose evidence survived verification
- the score must sit in the band its own verified verdicts imply — any misattribution caps it at 1, all-supported requires at least 2

A **supported** verdict's evidence must come from the block the occurrence cites. A **misattributed** verdict must name the block that actually supports the claim in `supporting_marker`, that block must differ from the one cited, the evidence must come from it, and the evidence must **not** also verify against the cited block. Retrieved blocks overlap routinely — sliding windows, shared boilerplate, two documents quoting one rule — so evidence present in both cannot show which of them supports the claim, and evidence the cited block contains shows that block supporting it. The prompt asks the judge for a quote unique to the supporting block, so a correct verdict is not lost to its choice of evidence.

A confirmed fabrication clamps the score to 0 in Python and leads the explanation, with the judge's own text following and preserved in `details["judge_explanation"]`. Where a fabrication is already established, an unusable judge reply does not rescue the row into un-assessed.

However a reply is rejected, the same record is kept: the judge's score under `rejected_raw_score`, its text, the number of verdicts discarded, and the verdicts that did verify. On an un-assessed row those go under `verified_verdicts` rather than `supported_citations`, because nothing there was assessed and a populated results key would read as though it had been.

## Assumptions

**Context format.** Retrieved context is a sequence of labelled blocks, each led by a bracketed source id at the start of a line and followed by a real boundary. This is what the reference RAG apps emit (`demo_app/finance_advisor.py:97`):

```python
retrieved = "\n\n".join(f"[{s['id']}] {s['text']}" for s in snippets)
```

Blocks with an empty body are excluded, since a source with no text can support nothing. Repeated labels are excluded as ambiguous source data rather than resolved to the first occurrence. Both remain labels the context declares, so a citation naming one is reported as ambiguous rather than treated as not a citation.

**Response format.** Citations appear as `[source-id]` or as a markdown link `[source-id](https://...)`. Both parse to the same marker, and the link flag is kept per occurrence, so a link form elsewhere in the response cannot mark a bare occurrence as link text. Matching is case-insensitive.

**Why this needed assuming.** There is no example of a cited response anywhere in the repo — the bundled datasets carry unlabelled context, and the demo apps call a live API so no cited output is checked in. The demo prompts say only "Cite the retrieved guidance when relevant" without specifying a syntax. The fixtures here are the first worked examples of the convention.

## Limitations

Most degrade to an **un-assessed row** rather than a wrong score, surfacing in the coverage-gap report so a reviewer sees the gap instead of trusting a number. The exception is a marker that names no source but does resemble the labels the context uses: that is a fabricated citation and fails outright.

| Limitation | Behaviour |
|---|---|
| Trailing-attribution context (`text. [id]`) | No blocks parsed → `unresolved_citations` |
| Unlabelled context | No blocks parsed → `unresolved_citations` |
| Label styles with no structure (`doc`, `1`) | Nothing of that shape is accused → `unsupported_label_style` when nothing resolves |
| Numbering mismatch (`[1]` against slug labels) | Structural mismatch → not an accusation |
| Bracket notation in prose or code (`arr[0]`, `[TODO]`) | Recognised as not being source ids → `ignored_markers`, no effect on the score |
| Document-level ids with chunk-level retrieval | Two chunks sharing `[doc-1]` make the label unusable → `partial_citation_coverage` |

`citation_pattern` and `source_label_pattern` are overridable for a different label **syntax** (covered by a test). They do not help with a different label **position**: block text runs from the end of one label to the start of the next, so a label must lead its block.

**Two deliberate misses.** An ambiguous marker may mean the retriever referenced a source without fetching it — in which case the response cited something not retrieved, arguably a real error that goes unreported. Because the alternative explanation (the parser missed a label) is also possible, no fabrication is claimed. Separately, a genuine misattribution whose evidence appears in both blocks is rejected rather than believed, so a real finding can be missed when the judge picks non-discriminating evidence. Both err towards a visible gap over a false accusation, which is the worse failure for a compliance report.

## Choices that go beyond #22

**Shape-gated fabrication.** The obvious implementation calls any unresolved marker fabricated. The label parser is not infallible, and a bracketed token may not be a citation attempt at all, so a marker is only accused when its signature matches how this context names its sources *and* it reads like a source id rather than an interval or coordinate pair.

**A deterministic fabrication floor.** This is the clearest divergence from the existing scorers, so it is flagged rather than left to be found. The established pattern is that the judge's score passes through unchanged and the judge is distrusted only about evidence, with unverifiable spans discarded and counted. This scorer extends that distrust to the score itself, but narrowly: the override fires only where Python has independent proof that a cited source is absent from the context, never because a score looks wrong. The judge's own verdict is preserved in `details` alongside `floor_applied`, so the override is auditable rather than silent. Happy to drop it if you would rather this scorer followed the existing pattern exactly.

**Validating the score rather than deriving it.** The round-2 review offered either. Deriving from outcomes would make rubric band 2 ("supported, with harmless imprecision") unreachable, since outcomes are only supported or misattributed, so the score is kept and checked against the band its verdicts imply.

**Ten branches in `assessment/assessor.py`.** `_classify_unassessed_reason` is a hardcoded allowlist; without registering these skip reasons they fall through to "see explanation in JSON report" and the coverage report loses the detail.

## Where the review left the decision open, and what I picked

Three places the last review offered a choice or stopped short of specifying the consequence. Say the word on any of them and I will switch.

**Fabrications and the coherence rule.** The options were to include known fabrications in the coherence rule, or to tell the judge to score only the resolved citations and let the floor own the fabricated result. I took the second: the rubric now scopes the judge to the citations it is given, and rubric band 0 no longer describes a citation naming an absent source, since the judge is no longer asked to score one.

**Rejecting a contradicted misattribution.** The instruction was to reject the verdict; what happens to the row afterwards was not specified. A rejected verdict is treated like any other unverified one, so the occurrence is left uncovered and the row is un-assessed. Rewriting it to `supported` would mean recording a verdict the judge did not return.

**`response_span` when the quote does not verify.** The options were to retain a verified `response_span` or to document a revised contract on #22. I retained it, and the key is **omitted** rather than emitted empty when the judge's quote does not verify against the response. `GroundednessScorer` always carries the field because it drops the whole pair when either half fails; here the verdict survives on its context evidence, so the field is sometimes absent. Flagging it because it is a small divergence from the established shape rather than a match.

## Not registered in `scorer_registry.py`

Following `GroundednessScorer`. `resolve_scorers` only instantiates registered classes, so registering would auto-run this on non-RAG rows and flood reports with un-assessed noise. Use via `additional_scorers`.

## Follow-ups worth filing

`demo_app/finance_advisor.py` already emits `"retrieved_ids": [s["id"] for s in snippets]` in `ModelResponse.metadata`, but `evaluation/pipeline.py` currently forwards only the flattened `retrieved_context` string, so the ids do not reach the scorer. If they were forwarded through `scorer_extras`, this scorer would not need to parse labels out of a string at all, and every parsing limitation above would disappear — including the chunked-retrieval case, where chunks would arrive as distinct units carrying their document id. The same plumbing would unblock the per-chunk metrics still open on #4.

Regex parsing here is a deliberate interim step, not the intended end state.

Separately, `judge_output_rejected` is recorded on assessed rows but nothing aggregates it, so a run cannot report how often the judge returned unusable output. That is a toolkit-wide observability question rather than something this scorer should answer alone.

## Testing

- `python -m pytest -q` — **226 passed, 4 skipped** (the skips are Weave tests; CI runs them in its own job), of which **157** cover this scorer
- `ruff check --select E9,F63,F7,F82 .` — clean
- `reuse --no-multiprocessing lint` — compliant
- `git diff --check` against the base — clean
- `ruff` on the changed files reports the same findings as the same files on `main` — no new ones

Rebased onto current `main`, so #46 is included.

Tests mock `_call_judge`, so no key or network is needed. Coverage includes every guard asserting the judge was not called, hyphenated, underscored, bare-word and numeric label styles, both orderings of the mixed link/bare form, empty and duplicate blocks, markdown links and reference definitions, malformed and non-object judge replies, contradictory verdicts, a score disagreeing with its verdicts, misattribution evidence from the cited block and from an unknown one, overlapping blocks in both directions, and the annotation round-trip on responses containing the tag characters — with controls that the legitimate version of each still scores.

Alongside the committed tests I ran a combination matrix over label style × block overlap × stray token × fabrication form × judge-reply validity, asserting properties rather than expected outputs, and an audit that checks every instruction in the generated prompt against what the verifier actually enforces. The matrix is not included here: the repo's tests are example-based throughout, and introducing a second testing idiom seemed worth raising separately rather than folding into this PR. Happy to open it as its own change if useful.

Each fix was verified by reverting it and confirming the tests fail, so none of them passes for an unrelated reason.

The judge prompt has still not been exercised against a live model. The deterministic behaviour is covered, and a live-key test was confirmed unnecessary for these fixes.

## AI assistance

This contribution was AI-assisted (Claude). The design decisions, the occurrence model, the shape-gated accusation rule and the deterministic floor were worked through and reviewed by me, and I can explain or rework any line.


knisar:
Thanks for this, it's a careful piece of work and the writeup made review easy. I checked out 8513562, ran the 24 new tests independently and they all pass. It follows what we agreed in #22. Requesting changes for four things that are specific to this PR:

1. The fabrication floor fires on things that aren't citations. Once one real citation resolves, any other bracketed token becomes "fabricated" and clamps the score to 0. I reproduced this with arr[0], [x], [TODO] and [Wikipedia](https://...) alongside a valid [adverse-action]. All four scored 0.0. The description says arr[0] lands in the safe path, but that only holds when nothing else resolves. Suggestion: only treat a marker as a fabrication candidate when it matches the shape of the labels that did resolve (slug vs numeric), and drop the [text](url) form unless the anchor itself resolves. Open to another approach, but please add regression tests for those four cases.

2. Evidence from uncited blocks is accepted. Response cites [source-a], judge returns a supported item tagged source-b, and it's accepted as verified because haystacks is built from every parsed block rather than the resolved set. Scored 1.0. Build the haystacks from resolved only and add a test showing evidence for an uncited marker gets discarded.

3. Ambiguous markers still reach the judge. Python is careful not to accuse on ambiguous markers, but the full response goes to a prompt that says grade every [source-id]. List the resolved markers in the prompt and tell the judge to grade only those. Pairs naturally with the fix for 2.

4. The all-unresolved branch throws away information. When nothing resolves, fabricated_citations comes back empty even though the markers were extracted. Please keep them in details regardless. And I'd go one step further: if blocks parsed successfully (so the context is labelled) and a marker passes the same candidate test as item 1, that's a fabricated citation and should fail, not skip. Structurally mismatched markers (numeric against slug labels) can stay un-assessed as a likely format mismatch.

Two things you flagged are real but predate this PR: LLMJudgeScorer.__init__ wiping category to "", and the Weave-native path not forwarding additional_scorers to get_detailed_evaluation. I'll open issues for both tracked in #28 and #29 so they don't block this.

Happy to approve once these land with tests.

M4h1m4:
Thanks @knisar , that's a thorough review. I've reproduced all four locally and agree with each.                                                                                                                       

On 1: you're right that the "at least one other marker resolved" guard makes this worse rather than better, since resolving a real citation is exactly what unlocks the false accusations. Going with the shape test, and dropping link text unless the anchor itself resolves. 4 uses the same mechanism, and matching the shape of the parsed labels is a better signal than the one I had.

I'll also correct the PR description, which claims arr[0] lands in the safe path. That only holds when nothing else resolves.

Will push with regression tests for the four cases and get back to you.

M4h1m4:
Hi @knisar All four items are fixed, pushed as three commits on top so you can read them separately.
  
1 and 4 turned out to be one mechanism. Markers are now compared against the shape of the labels the context actually uses, and only a marker matching that shape can be called a fabrication. That gates both the floor and the fail-vs-skip decision. Your four cases all pass now, with a control confirming a real fabrication alongside stray brackets still fails.
  
One deviation: slug vs numeric as worded fixes arr[0], and the link rule fixes [Wikipedia], but [x] and [TODO] are alphabetic and still read as slug-shaped, so two of the four survived. I widened the signature to (is_numeric, has_hyphen, has_dot), which covers all four and still flags reg-z-2024. Happy to narrow it if you'd rather keep the simpler rule.
  
2 and 3 are in one commit since they're the same idea from either end. Worth noting 3 wasn't redundant with 2: the judge returns a single holistic score, so a marker it graded and disliked drags the score down even after its span is discarded.
  
I also fixed the arr[0] claim in the description you called out, and the docstring, which had the same error.

knisar (changes requested):
Thanks for the careful follow-up. I went through all three corrective commits at `e2818d0` and reran the full suite in both environments. The four points from my first pass are fixed: evidence is scoped to cited blocks, ambiguous markers stay out of the judge prompt, unresolved rows keep their marker details, and the original stray-bracket cases behave as intended. Both runs finish at 58 passed.

I still cannot approve this yet because the scorer can return confident compliance results for citations it has not actually established.

1. The marker-shape check is still too coarse for label styles the class claims to support. With context labelled `[doc]`, both `f([x])` and `[TODO]` have the same signature as the real source ID and are classified as fabricated. With context labelled `[1]`, `arr[0]` is classified as fabricated. In all three cases I gave the judge a 3/pass response and the deterministic floor changed it to 0/fail. The current regressions only use hyphenated labels, so they do not catch this. Classification is also order-dependent: `[fake-id](url)` followed by `[fake-id]` is ignored, while the reverse order is treated as fabricated because extraction keeps the first form it sees. Please add alphabetic, numeric, and mixed-form tests, then make the default conservative when the source ID grammar cannot distinguish citations from ordinary brackets. A stricter configured pattern or an explicitly unsupported and un-assessed label style is safer than a false compliance failure.

2. The judge result needs validation before it becomes an assessed score. A judge score of `"NaN"` currently normalizes to 1.0 and passes. A score of 3 for two resolved citations also passes when the returned evidence covers only one of them. Please require a finite numeric score in the 0 to 3 range and a valid supported or misattributed verdict for every resolved marker. Invalid or incomplete judge output should be un-assessed unless a deterministic local finding already proves failure.

3. Partial resolution is reported as a complete assessment. I tested a response citing `[source-a]` and `[source-b]` where `source-a` is a block label and `source-b` only appears inside that block. The judge sees and passes `source-a`; the final result is 1.0, passed, and assessed even though `source-b` remains in `ambiguous_citations`. With the current binary `assessed` field, any unresolved citation should make the row un-assessed. The alternative is an explicit partial state that cannot be aggregated as complete.

4. The deterministic floor can contradict the user-facing explanation. I reproduced score 0 and `passed=False` with the explanation `The real citation is fully supported.` Keep the judge text in details if it is useful for audit, but the result explanation needs to include the deterministic fabrication finding. Please assert the final explanation in the floor regression.

5. Empty and duplicate source blocks should not be treated as valid. `[source-a]` immediately followed by `[source-b] Other text` can receive a full pass for the empty `source-a` block. Repeating `[dup]` keeps the first block silently while the judge sees both blocks, so evidence from the second is discarded but the raw score can still pass. Filter empty blocks and treat duplicate labels as ambiguous source data rather than choosing the first one.

The core integration itself checked out. The scorer receives the model's retrieved context, `expected`, and `rubrics`; it preserves category `MIT-3.1`; and its verified evidence survives the pipeline. A live-key test is not needed for these deterministic fixes.

Please also refresh the branch onto current main when you push the next update. #33 has landed, and the category workaround described in the PR body is stale now that #30 is merged. #31 remains the separate Weave handoff dependency.

Zeming-Yuan:
Hi! I'm the author of #37 (`RetrievalRelevanceScorer`) — knisar asked both
PRs to settle on one context contract before they land.

I've aligned #37 to the same labelled-block convention you use here:
line-start `[source-id]` blocks parsed with the same character set as
`_SOURCE_LABEL_PATTERN` (labels only count at line start, so bracketed text
inside a passage never splits a block). #37 keeps a `---`-delimiter fallback
for unlabeled contexts, since retrieval relevance is also meaningful on raw
text, but native labelled contexts take precedence.

If you'd rather share a single parsing helper in a follow-up (e.g. hoisting
`_SOURCE_LABEL_PATTERN`/`_parse_source_blocks` into a common module both
scorers import), I'm happy to rebase onto that once one of the two PRs
merges. Flagging here so neither PR fixes the contract twice.

M4h1m4:
Hi @Zeming-Yuan, thanks for flagging this, and for aligning #37 to the labelled-block convention. Agreed on settling one contract, and I am happy to hoist _SOURCE_LABEL_PATTERN and _parse_source_blocks into a
  shared module once either PR lands.

One heads-up before you build on the current parsing. @knisar's second review here asks for changes to _parse_source_blocks itself: empty blocks will be filtered rather than treated as valid sources, and duplicate labels will be treated as ambiguous source data instead of silently keeping the first. So the helper's behaviour is about to change even though its signature and the label grammar stay the same. I will comment here once that is pushed so you can see the final shape before rebasing onto it.

The ---- fallback for unlabelled context makes sense for retrieval relevance, since that metric is still meaningful on raw text. This scorer goes the other way and returns un-assessed when it cannot parse labels, because without them it cannot tell a fabricated citation from a citation-format mismatch. That reads to me as different calls for different metrics rather than a contract disagreement, but say if you see it otherwise.

M4h1m4:
Thanks @knisar , this was a thorough second pass. All six points are fixed, pushed as three commits on top of the rebase so you can read them separately. Rebased onto current main (through #48), and the category note you
flagged is gone from the description now that #30 is merged.

Items 1 and 4 turned out to share a mechanism, so they are in one commit. A grammar check now runs before any accusation: source ids only support the distinction when they carry a separator, so labels must contain -, . or _. Where they do not, no marker can be a fabrication candidate, and a row with nothing resolved is un-assessed under unsupported_label_style. All three of your label-style reproductions now pass at 1.0 with the stray marker recorded under ignored_markers, and a control confirms hyphenated labels still flag reg-z-2024.

This does lose genuine fabrications under bare-word and bare-number schemes, which I took to be the trade you were asking for. Two notes on how I read your suggestion:

- slug vs numeric as literally worded fixes arr[0], and the link-text rule fixes [Wikipedia](url), but [x] and [TODO] are alphabetic and still read as slug-shaped, so two of the four cases survived. Hence the separator test rather than a numeric one.

- You offered a stricter configured pattern or an unsupported and un-assessed label style. I took the second as the default so the scorer keeps working on the format the demo apps emit, with citation_pattern available for anyone who wants accusation on a bare-word scheme. Happy to invert that if you would rather it never accuse without explicit configuration.

 Order dependence is gone. The link flag is collapsed across every occurrence rather than taken from the first, so a marker written as a link anywhere is treated as a link throughout. I took the conservative direction on the grounds that mixed forms are weak evidence of a citation attempt.

Judge output is validated before it becomes a score. Finite, numeric, within the rubric range. Two cases you did not hit were worse than the ones you did: "abc" and a missing score key both raised out of float() as unhandled exceptions rather than producing a wrong number. Validation runs before the normaliser rather than inside it, so this stays independent of #50 and does not turn into a traceback when that lands.

Coverage is measured against the verified verdicts, not the raw reply. A verdict whose evidence could not be found in the row is discarded, and a discarded verdict has established nothing, so counting it would let invented evidence satisfy the requirement. A misattributed verdict counts as covered, since the question is whether every citation got a checkable answer rather than a favourable one. Invalid or incomplete output is un-assessed under invalid_judge_score or incomplete_judge_verdicts, kept separate because one points at the judge model and the other at the prompt. Where a fabrication is already established the row still fails rather than becoming un-assessed.

Partial resolution now makes the row un-assessed under partial_citation_coverage. I took your first option rather than adding a partial state, since that would change ScorerResult and the aggregation path for every scorer. One reading to flag: I applied this to ambiguous markers only, not to ignored ones. An ignored marker was determined not to be a citation, so treating it as blocking would send any response containing an array index back to un-assessed and undo the stray-bracket fix. There is a control test for that.

The floor now explains itself. It leads with the deterministic finding and names the absent sources, with the judge's tex following rather than dropped, and the unmodified judge text kept in details.judge_explanation. The floor regression asserts the final explanation as you asked.

127 tests, ruff and reuse clean, and git diff --check clean against the new quality gates.

M4h1m4:
@Zeming-Yuan - the parsing changes I flagged earlier are now in. _parse_source_blocks keeps the same signature and label grammar, but excludes two things it used to accept: blocks whose body is empty, and repeated labels, which are now treated as ambiguous source data rather than resolving to the first occurrence. If #37 lands first I am happy to rebase onto a shared helper; if this one lands first the helper is ready to hoist as it stands.

knisar (changes requested):
Thanks for the detailed update. I rechecked the exact head at `9dbf012` against current main after #51 landed. The merge is clean. All 74 citation-focused tests and all 140 repository tests pass, and the configured Ruff, REUSE, and whitespace checks are clean. The items from my previous review have moved in the right direction, but I can still reproduce five correctness problems that need to be resolved before approval.

1. Citation occurrences are collapsed by marker. If two different claims both cite `[source-a]`, one supported and one unsupported, a single verified verdict for the supported occurrence is enough for a 1.0 assessed pass. I also swapped the response spans for `[source-a]` and `[source-b]`; both associations were wrong, but the result still passed because each span only has to occur somewhere in the response. Please preserve citation occurrences or their attached claim spans, and require a coherent verified verdict for every cited claim occurrence.

2. A Markdown link at the start of a line is parsed as a source block. With `[docs](https://example.com) This is a Markdown link` in the context, the parser creates a `docs` source whose body starts with the URL. A response citing `[docs]` can then receive a full pass. The source label needs a real boundary after `]`, while `](` must be excluded. Please add Markdown link and reference-definition regressions, and keep this boundary consistent with #37.

3. The label grammar still gives both false failures and false passes. `_labels_are_distinguishable()` treats `_` as structure, but `_marker_signature()` does not, so `[source_one]` plus an ordinary `[TODO]` token is floored to 0. In the other direction, a context containing `[doc-1]` and `[2]` lets a missing `[doc-99]` be ignored and still return a 1.0 assessed pass. I also found that `arr[0]` copied from the context becomes ambiguous before its shape is checked. Please make the grammar decision per relevant label style and apply the shape check before inline-presence handling. The regressions should cover underscore labels, mixed label styles, and normal bracket notation copied from source text.

4. Judge verdict coherence and misattribution semantics are not enforced. The same marker can appear in both the supported and misattributed lists and still pass. A verified misattribution combined with judge score 3 also returns 1.0 and `passed=True`. In addition, evidence from the cited block itself is accepted as proof that the citation was misattributed. Please require exactly one verified outcome per citation occurrence, reject contradictory verdicts, and either derive the final result from those verified outcomes or validate that the judge score agrees with them. Misattribution evidence must establish support from a different source block, ideally by carrying and validating that source marker.

5. Judge response validation is still incomplete. JSON booleans are accepted as numeric scores because `float(True)` succeeds. Valid JSON with a non-object top level, such as `null` or `[]`, raises instead of returning a controlled unassessed result. In the deterministic fabrication path, `NaN` and infinities can also remain in `details.raw_score`, which makes strict JSON serialization fail. Please reject booleans, handle non-object replies as invalid judge output, and keep rejected raw values JSON-safe. This should use the same serialization contract requested on #50.

There are also three small cleanup items for the next push. Please rebase onto current main so #51 is included. Change the opening line to `Closes #22. Part of #4.` so the issue closes when this merges. The PR body still contains the stale category workaround even though #30 fixed it, and the validation counts are now out of date, so please refresh those as well.

Once these are pushed, I will rerun the focused adversarial cases and the full integration suite.

M4h1m4:
@knisar All five fixed. Cleanup done: rebased onto current main (#51 included), opening line is now Closes #22, and the description is rewritten rather than patched since most of it described the pre-round-3 design.
  
Three decisions worth your eye:

Claim spans — I took a third path. Deriving the claim by slicing between markers turned out fragile on ordinary prose: The rule [x] requires notices graded only "The rule", and a citation opening a sentence produced nothing. Preserving the judge's span instead would have kept the association checkable but not exact. So the response now reaches the judge with an occurrence tag after each gradeable citation:

The rule [adverse-action]⟦1⟧ requires notices, and deadlines apply [fair-lending]⟦2⟧.

The judge reads the prose as written and answers per number. Nothing here decides where a claim begins or ends, and binding stays exact because the number is attached to the marker. claim_span is advisory:
verified against the response, discarded otherwise, never binds a verdict.

Shape check ordering. Applied before inline-presence handling as you asked, but only when a structured style exists to compare against. Unconditionally it collapsed the ambiguous tier for duplicate and empty labels, where there is no style to judge shape by and a bracketed marker is still a plausible citation.

Score - validated rather than derived, of the two you offered. Deriving would make rubric band 2 unreachable, since outcomes are only supported or misattributed, and that band distinguishes a clean 3 from an
  imprecise 2.

Smaller calls: added the underscore bit to the signature rather than dropping _ from the separator list, since removing it would make underscore-labelled contexts entirely un-accusable; renamed the skip reason to invalid_judge_output since it now covers replies that are not scores at all. Rejected values follow the #50 contract; I have not touched the normaliser.

I swept for the same class before pushing and found seven more, including occurrence: true being accepted — the same bool-subclasses-int defect as the score, one field over. All eight fields read from the reply now have type guards.
  
182 tests, 4 skipped for Weave. Ruff, REUSE and git diff --check clean — run locally, since CI still needs maintainer approval.                                                                               

@Zeming-Yuan — _parse_source_blocks changed again: a label now needs a real boundary after ], so Markdown links and reference definitions at line start no longer parse as sources. Character class and eading-label rule unchanged.

knisar (changes requested):
Thanks for the careful follow-up. I rechecked the exact head at ff72ceb against current main at 4780f9b. The five items from my previous review are fixed in the cases I originally reported. The merge is clean, all 122 focused tests and all 188 repository tests pass locally, and the configured Ruff, REUSE, dependency, package build, and whitespace checks pass.

I found five deterministic cases that still need correction before approval.

1. Mixed bare and link forms can launder a fabricated citation. With context [doc-1] Real source text, a response containing a valid [doc-1], a bare [doc-99], and [doc-99](https://example.com) returns assessed=true, score=1.0, passed=true. The link form makes every occurrence of doc-99 look like link text, so the bare fabrication is ignored. Keep link state per occurrence, or make a mixed form ambiguous, but a link occurrence must not erase a bare occurrence.

2. Dropped empty and duplicate labels are lost when another valid block uses a different label shape. With two [dup] blocks plus a valid [keep-this] block, a response citing both is assessed at 1.0 because dup is put in ignored_markers. The same happens with an empty [empty] block followed by a valid [doc-1] block. Preserve invalid line-leading labels separately and classify citations to them as ambiguous before applying general shape filtering.

3. The fabrication score contract contradicts the prompt. With one supported [doc-1] citation plus fabricated [doc-99], the prompt tells the judge to factor the absent source into the score. A correct score of 0 with a supported verdict for doc-1 is then rejected as judge_score_contradicts_verdicts, and the verified support details are discarded. Scores 2 and 3 are treated as coherent before the local floor. Either include known fabrications in the coherence rule, or tell the judge to score only the resolved citations and let the deterministic floor own the fabricated result.

4. Duplicate evidence across source blocks can create a false misattribution. If [doc-1] and [doc-2] both contain the exact text The policy requires notice, a response citing doc-1 is accepted as misattributed when the judge names doc-2 and quotes that text. The result is an assessed failure even though the cited block visibly contains the same evidence. Reject a misattributed verdict when its evidence also verifies against the cited block.

5. Occurrence tag sanitization can change the claim sent to the judge. The response The rate is ⟦1⟧5% [doc-1] reaches the prompt as The rate is 15% [doc-1]⟦1⟧. Removing the tag characters while keeping their contents changes the value being judged. Please use an annotation token that is absent from the row, or escape collisions without changing the response text.

Please also reconcile the evidence output with #22. The agreed shape was verified response_span and context_span fields plus the citation marker. This head returns claim instead, and an invented or swapped claim can become empty while the verdict remains assessed. No current renderer crashes, but downstream consumers lose the established response evidence field. Retain a verified response_span alongside the occurrence metadata, or explicitly agree and document the revised contract on #22.

The current branch is much closer, and the basic occurrence numbering, completeness checks, contradictory-verdict checks, score type validation, and integration path all behaved correctly.

knisar:
One related classification case surfaced in the same sweep and is worth covering in this pass. With [doc-1] as the real source label, the response The normalized interval is [0-1]. Valid claim [doc-1]. is floored from an otherwise valid pass to 0. [x-y] and a source-shaped token inside a fenced code block behave the same way. The separator signature establishes syntactic similarity, but not that the bracket is a citation. Please include this in the classification fix by narrowing when an unresolved bracket is accusation-grade, excluding ordinary notation and code, or returning un-assessed when citation intent cannot be established.

knisar:
One more direct path belongs with item 3. With a valid [doc-1] citation plus confirmed fabricated [doc-99], a judge reply of valid JSON null or [] returns assessed=false with invalid_judge_output. That turns a deterministic failure into a coverage gap, contrary to the stated guarantee that unusable judge output cannot rescue a proven fabrication. Please apply the fabrication failure before returning from the non-object guard and add both null and list regressions.

M4h1m4:
Thanks @knisar  — all seven are fixed at `8aac17a`, rebased onto `4286a16`. PR description refreshed.

| Item | What changed |
|---|---|
| Mixed bare/link launders a fabrication | Link state is kept per occurrence, so a link form cannot mark a bare occurrence as link text |
| Dropped empty/duplicate labels lost | `_parse_source_blocks` now returns the labels it rejected, and resolution checks that set **before** the shape filter, so a citation naming one is ambiguous |
| Fabrication contract contradicts the prompt | Took your second option: the judge is scoped to the resolved citations and the floor owns the fabricated result. Rubric band 0 no longer describes scoring a citation that names an absent source — it describes misattribution of the central claim, and the scoping rule sits in the rubric itself rather than only in an appended block |
| Duplicate evidence across blocks | A misattributed verdict is rejected when its evidence also verifies against the cited block. The prompt now asks for a quote unique to the supporting block, so a correct verdict is not lost to its choice of evidence |
| Occurrence tag sanitization | The delimiter pair is chosen per response from a candidate list, falling back to private-use codepoints. Nothing is stripped, and removing the inserted tags returns the response verbatim |
| `claim` vs `response_span` (#22) | Renamed, and omitted rather than emptied when the quote does not verify |
| `[0-1]`, `[x-y]`, code-fenced tokens | A signature match is no longer sufficient — the marker must also carry a part that reads as a word — and code spans are excluded from extraction |

Three more fixed alongside them:

- A `null` or `[]` reply hit a non-object guard that bypassed the fabrication path, so an unusable reply could turn a proven fabrication into a coverage gap. Every rejection now routes through one place.
- A rejected reply reports the judge's score, its text, the number of verdicts discarded and the verdicts that did verify, whichever branch it takes. Previously that record was kept only when a fabrication was present, so a verdict that verified could be dropped because its neighbour failed. On an un-assessed row those go under `verified_verdicts` rather than `supported_citations`, since nothing there was assessed.
- The occurrence-tag constraint — that a candidate pair can never be `[` or `]` — is asserted in a test rather than left in a comment.

Three places the review left the choice open, so flagging what I picked:

- **After rejecting a contradicted misattribution**, the verdict is treated like any other unverified one: the occurrence is left uncovered and the row is un-assessed. Rewriting it to `supported` would mean recording a verdict you did not return.
- **`response_span` is absent, not empty**, when the quote does not verify. `GroundednessScorer` always carries the field because it drops the pair; here the verdict survives on its context evidence, so this is a small divergence from that shape.
- **Band 0** describing misattribution follows from taking your second option on fabrications — it is the only thing left for it to mean once fabrications are scored deterministically.

One known trade: a genuine misattribution whose evidence appears in both blocks is rejected rather than believed, so a real finding can be missed when the judge picks non-discriminating evidence. The prompt change reduces it; I would rather miss that than fail a correct citation.

226 tests pass, 157 on this scorer. Each fix was checked by reverting it and confirming the tests fail. I also ran a combination matrix over label style × block overlap × stray token × fabrication form × judge-reply validity, and an audit checking every instruction in the generated prompt against what the verifier enforces — both at zero findings. Neither is in the PR; the matrix would introduce a second testing idiom, which felt like your call.

Two I have not acted on: `judge_output_rejected` is recorded per row but nothing aggregates it, so a run cannot report how often the judge returned unusable output; and the `retrieved_ids` plumbing noted in the description would remove the label parsing entirely, including the case where chunk-level retrieval reuses a document id.

Happy to rework any of it.

knisar:
Thanks for the detailed update. I reran 8aac17a and found three cases that are still merge blockers.

1. The source-label and citation patterns both use `\s*`, so brackets can span line breaks. `[\ndoc-1]` is accepted as a source label, and `[\ndoc-99\n]` can be extracted as a citation. Please use horizontal whitespace around IDs and cover both paths.

2. The code-span exclusion handles single backticks and closed triple-backtick fences, but misses tilde fences, double-backtick spans, and an unclosed fenced block. In each case `[doc-99]` inside code is treated as a fabricated citation and floors an otherwise valid result. Please cover these standard Markdown code forms.

3. `_occurrence_tag()` can exhaust its candidates. A finite response containing the preferred delimiters and the private-use fallback characters raises `IndexError` before the judge call. Please generate a collision-free variable-length tag, or return a controlled unassessed result if none can be constructed.

#37 is now merged, and this branch conflicts with it in `README.md` and `rai_toolkit/scorers/llm_judges.py`. Please rebase onto current main and resolve that shared parser area as part of this update. The existing 157 focused tests and 226 full tests pass on 8aac17a, so the remaining work is limited to these boundary cases and the rebase.

M4h1m4:
@knisar All four are fixed at `b827938`, rebased onto `6d0ed0c`.

| Item | What changed |
|---|---|
| Brackets spanning line breaks | Both patterns use `[ \t]*` now. The rebase brought #37's source-label pattern, which already had the fix; the citation pattern needed the same. `\s` also matches carriage return, form feed, vertical tab and the Unicode separators, so those are covered too |
| Missing code forms | Fences open with three or more backticks **or** tildes and close on a run of the same character. One opening a line may also run to the end of the response, since a truncated block is still code; one opened mid-line must close, or a stray run in prose would hide every citation after it. Inline spans close on a run of the same length as the opener, so `` ``x`` `` is one span rather than two empty ones either side of `x` |
| `_occurrence_tag()` exhaustion | Single characters are a preference now, not the whole search. Once the named pairs and the private-use range are spent the delimiters lengthen by repetition, which cannot be exhausted. I took the first of your two options because returning un-assessed would lose real citation coverage whenever a response happened to contain unusual characters. The caller still handles a `None` tag with a registered skip reason, since this defect came from trusting an invariant rather than checking it |
| Rebase and the shared parser | There is one `_SOURCE_LABEL_PATTERN` now, #37's. My own commit had added the same boundary assertion independently, so I dropped it rather than keep two competing definitions — `_parse_source_blocks` takes #37's pattern as its default, so both scorers read the context format through the same regex |

One more, found while re-running the prompt audit after the tag change: the tag block told the judge the delimiter characters appear nowhere else in the response. That holds for a single-character tag and stops holding for a lengthened one, where the characters appear throughout and only the sequence is absent. It was false exactly where it does the most work, so it now claims what is actually guaranteed.

327 tests pass, 214 of them on this scorer, and #37, the Anthropic adapter and the prompting tests are green on the rebased head. `ruff --select E9,F63,F7,F82`, `reuse lint` and `git diff --check` are clean, and `ruff` over the whole tree reports the same 107 findings as `main`. Each fix was checked by reverting it and confirming the tests fail.

On the number of rounds: five is more than this should have taken, and the review load has been mine to reduce rather than yours to absorb. Two of the round-4 items were half-fixes of your own earlier findings — I made the case you reported pass without removing the thing it contradicted - which is why round 5 existed at all. What changed since is that I now audit the generated prompt against what the verifier enforces, sentence by sentence, instead of re-testing the scenario you named; that is what caught the tag wording above before you saw it. Thank you for the precision in each round, it made every item unambiguous to reproduce.

Happy to rework any of it.

knisar (changes requested):
Thanks for the update. I rechecked head `b827938`. The latest requested fixes are working, and the existing validation is green with 215 focused tests passing and the full suite completing with 367 passed and 4 skipped.

I still reproduced three citation parsing issues:

1. A fenced code block opened with three backticks can validly close with four or more backticks. The current parser does not recognize the longer closer, so a later unsupported citation can remain hidden inside the masked region and incorrectly receive a full pass. Please cover longer closing runs for both backtick and tilde fences.
2. Markdown code spans can cross a line ending. Citation-like text inside a multiline code span is currently parsed as a real citation and can incorrectly produce a fabrication failure. Please cover single-backtick and multi-backtick multiline spans.
3. Custom source-label patterns do not preserve duplicate or empty declared labels during resolution. With a custom `<<source-id>>` pattern, an ambiguous duplicated label can be treated as valid and receive a full pass. Please use the selected source-label pattern when collecting rejected labels, and add duplicate and empty-block cases for custom syntax.

Please address these cases before the next review round.

M4h1m4:
@knisar All three are fixed at `e5c98cd`, rebased onto `e3186d6`.

| Item | What changed |
|---|---|
| Longer closing runs | A fence now closes on a run of the same character at least as long as the opener, for both backticks and tildes. A shorter run still does not close a longer fence |
| Multiline code spans | Inline spans may cross a line ending, with the closing run still matching the opener's length |
| Custom source-label patterns | The rejected-label lookup uses the scorer's own label pattern, so duplicate and empty declared labels under `<<source-id>>` behave exactly as they do under `[source-id]` |

One thing I left alone, in case you would rather it stayed as is: `_LABEL_SEPARATORS` is module-level while the patterns are overridable, so a custom syntax using a separator it cannot read makes no style accusable and a fabrication is missed rather than accused. That is the same trade bare-word and bare-number styles already make. Making it a class attribute alongside the two patterns would close it.

369 tests pass, 237 of them on this scorer, and #37, the Anthropic adapter and the prompting tests are green on the rebased head. Verified on Python 3.10. Each fix was checked by reverting it. Ruff, REUSE and the diff check are clean.

knisar (changes requested):
Thanks for addressing the three items from the last review. I verified the longer fence closers, multiline code spans, and custom duplicate and empty labels at `e5c98cd`. The proposed merge with current `main` is clean, and the local test, licensing, lint, and whitespace checks pass.

I still reproduced four masking boundaries. In each case, a real unsupported `[reg-z-2024]` citation disappears from extraction after a valid `[adverse-action]` citation, and the scorer returns `assessed=True`, `score=1.0`, and `passed=True`.

1. CRLF fenced blocks are not closed because the scanner removes `\n` but leaves `\r` on the closing line. The mask then extends to the end of the response.
2. A backtick fence opener whose info text contains a backtick is treated as a fence even though that is not a valid Markdown fence opener.
3. A fence marker indented four spaces is treated as fenced code. Fenced blocks allow at most three leading spaces, so later unindented citations should remain visible.
4. A one-backtick inline span is closed by one tick taken from a later two-backtick run. Inline delimiters must use matching run lengths, so the citation inside that unmatched span should remain visible.

Please fix these boundaries and add regressions that assert the unsupported citation remains visible and triggers the fabrication failure.

I am comfortable leaving `_LABEL_SEPARATORS` unchanged in this PR. Arbitrary custom separator syntax is outside the contract in #22. If we want to make separator rules configurable later, we can scope that separately.

M4h1m4:
All four are fixed at `9a58231` @knisar. 

| Item | What changed |
|---|---|
| CRLF fenced blocks | Line bodies are taken with `rstrip("\r\n")`. The stray `\r` had been failing the closing fence's end-of-string match |
| Backtick in the fence info text | A backtick fence's info string may not contain a backtick, so `` ```a`b `` opens no block. Tilde fences still may |
| Fence indented four spaces | At most three leading spaces open a fence; a tab counts as four columns |
| Inline run pairing | Both runs must be maximal, so a one-backtick opener can no longer close on the first tick of a longer run |

Each case is covered by a regression asserting the citation stays visible and that the row fails with `fabricated_citations: ["reg-z-2024"]`.

Three more over-masking cases turned up while checking these, all the same direction:

- A fence's info string is a language tag rather than content, so the span starts at the line after the opening fence. The opening line is still a block boundary an inline span cannot reach across, so the mask extent and the barrier extent are now tracked separately.
- A code span cannot cross a blank line, which ends the paragraph it opened in. Two unrelated backticks in different paragraphs were pairing.
- A run opened mid-line no longer opens a fence. That was an extension of mine from an earlier round and it masked prose no renderer treats as code.

The masking is now checked differentially against a reference CommonMark implementation instead of against rules chosen by hand — that is what turned up the three above. Over generated documents it disagrees on roughly three in four thousand in the over-masking direction, all artefacts of indented code blocks, which stay unimplemented: four-space indentation also marks list continuations, so excluding it without tracking list context would silently stop grading real citations.

489 tests pass, 257 on this scorer, verified on Python 3.10. Ruff, REUSE and the diff check are clean.

knisar (changes requested):
Thanks for the careful update. I rechecked the new head against current main. The four cases from my previous review are fixed, the focused citation tests and full suite pass locally, and the hosted checks are green.

I found three remaining CommonMark boundary cases in the new masking code that should be addressed together:

1. `_INLINE_CODE_PATTERN` recognizes blank lines made with LF and CRLF, but not two bare CR line endings. Bare CR is a valid line ending in CommonMark. With ``Valid [adverse-action]. `x\r\ry [reg-z-2024]` z``, the unsupported marker is still masked and the scorer reports a pass. Please make the blank-line expression accept CR, LF, and CRLF independently, and add a regression showing that the later citation remains visible and fails.
2. The opening fence info string is metadata, not response prose. A token such as `[reg-z-2024]` in the info string should not be extracted as a citation.
3. `_FENCE_CLOSE` currently permits unlimited spaces or tabs before a closer. CommonMark permits at most three leading spaces. A four-space or tab-indented pseudo-closer must not end the fenced block. Please add regressions for that boundary.

These are all within the Markdown masking path changed in this update. Once they are covered, I expect this review to be complete. Thanks again for staying with the detailed edge cases.

M4h1m4:
All three are fixed at `e8430a1`, rebased onto `ec81e80`. @knisar 

| Item | What changed |
|---|---|
| Bare CR blank lines | The blank-line rule accepts CR, LF and CRLF independently, and treats CRLF as one ending rather than a CR followed by an LF |
| Fence info string | Masked again, so a token there is not extracted. A renderer puts it outside `<code>` since it is a language tag, but it is metadata either way rather than prose the model wrote. This is now the one place the masking deliberately differs from a renderer |
| Over-indented closer | A closer carries at most three leading spaces, the same bound as an opener. Four spaces or a tab no longer ends the block |

Two more turned up while covering these, both in the same over-masking direction:

- A fence indented four or more spaces was still opening a block when it closed. The indent bound had only been consulted for the unclosed case.
- Indented code blocks are recognised now. I had left them out for fear of masking list continuations, but the rule that prevents that is part of the format: an indented block cannot interrupt a paragraph, so it begins only where none is open, and a continuation line inside a list is never read as code. Leaving them out was also costing the other direction, since backtick runs either side of an indented line were pairing across it.

The masking is enumerated exhaustively against a reference CommonMark implementation now rather than sampled: every document of N lines from an alphabet carrying both sides of each boundary, crossed with all three line endings. That is what found the two above. Clean at 336 and 21,168 documents; at 1,185,408 it disagrees on 81, all documents opening with a failed backtick run where the reference also disagrees with its own behaviour on the same fragment in isolation.

Both gaps that produced this round are closed at the source: the fuzzer's alphabet had no bare CR, and I had checked the opening fence's indent bound without its mirror on the closer. The combination matrix also has a line-ending axis now, so every guarantee is crossed with CR and CRLF rather than LF alone.

543 tests pass, 281 on this scorer, verified on Python 3.10. Ruff, REUSE and the diff check are clean.

rai_toolkit/scorers/llm_judges.py
```diff
@@ -824,6 +829,1232 @@ def score(
...
+    paragraph_open = False
+    while index < len(lines):
+        body = _line_body(lines[index])
+
+        if not body.strip():
+            paragraph_open = False
+            index += 1
+            continue
+
+        if not paragraph_open and len(_leading_space(body).expandtabs(4)) >= 4:
```
knisar: [P1] Keep list continuation prose out of the code mask

A blank line resets `paragraph_open`, but it does not close a list item. Four spaces can therefore indent a second paragraph within that list rather than a code block. For example, using the existing test helpers:

```python
output = (
    "1. Notices are required [adverse-action].\n\n"
    "    Rates are capped [reg-z-2024]."
)
result = _covering_scorer(output, CONTEXT).score(output, context=CONTEXT)
```

At this head, extraction returns only `adverse-action`, and the result is `assessed=True`, `score=1.0`, `passed=True`, with no fabricated citations. The second paragraph is ordinary list prose under [CommonMark's list rules](https://spec.commonmark.org/0.31.2/#list-items), so `reg-z-2024` must remain visible and fail the row. The same failure occurs with a bullet list and with a nested `    - Rates ...` item after the blank line. Please interpret indentation relative to the active list container and cover these cases across LF, CR, and CRLF, retaining controls that actual indented code is still excluded.

knisar (changes requested):
The three items from my previous review are fixed at e8430a1, and all 281 citation tests pass. I also checked the proposed integration with current main at fe9c020 after resolving the README and llm_judges.py merge conflicts locally: 616 tests passed and 3 skipped, with clean syntax/undefined-name lint, REUSE licensing, and patch whitespace checks. Those temporary conflict resolutions have not been pushed to your branch.

The newly added indented-code handling introduces one blocking regression: list continuation paragraphs and nested list items after a blank line are masked as code, so a real fabricated citation disappears and the scorer passes the row. The inline comment includes a minimal reproduction. An independent reference CommonMark parser confirms those spans are prose. I reproduced this for ordered lists, bullet lists, and nested lists across LF, CR, and CRLF: all 9 regression assertions fail, while 6 genuine indented-code controls pass.

Please preserve the surrounding list structure when deciding whether indentation denotes code, add these regressions and controls, and rebase or merge current main to resolve the integration conflicts. This is confined to the indented-code path introduced in the latest commit; the prior three fixes do not need to be reworked.

rai_toolkit/scorers/llm_judges.py
```diff
@@ -1661,6 +1666,1288 @@ def score(
...
+            paragraph_open = False
+            index += 1
+            continue
+
+        indent = _indent_width(body)
+        # A line shallower than the open item has left it, and any item that
+        # one was nested in.
+        while containers and indent < containers[-1]:
+            containers.pop()
+        content_column = containers[-1] if containers else 0
```
knisar, line 1984: [P1] Preserve list containers across lazy paragraph continuations

A line with less indentation is not sufficient to close a list item: CommonMark permits an unindented continuation of its open paragraph. For example:

```python
output = (
    "1. Notices are required [adverse-action].\n"
    "More explanation.\n\n"
    "    Rates are capped [reg-z-2024]."
)
result = _covering_scorer(output, CONTEXT).score(output, context=CONTEXT)
```

The renderer keeps both paragraphs inside the ordered list item, but this loop pops the container on `More explanation.`. After the blank line, the four-space continuation is then masked as top-level code. Extraction loses `reg-z-2024` and the scorer returns `assessed=True`, `score=1.0`, `passed=True`, with no fabricated citations. Keep the container open for paragraph-continuation text before deciding that indentation closes it. Cover bullet and ordered lists across LF, CR, and CRLF; the missing source must remain visible and fail the row. See the [CommonMark list-item laziness rule](https://spec.commonmark.org/0.31.2/#list-items).

rai_toolkit/scorers/llm_judges.py
```diff
@@ -1661,6 +1666,1288 @@ def score(
...
+        if opened is not None:
+            containers.append(opened)
+            paragraph_open = True
+            # A list item begins a new block, so an inline span cannot reach
+            # across it. Recorded with no extent: it masks nothing, it only
+            # stops a backtick run before the item pairing with one after.
+            start = offsets[index]
+            spans.append((start, start, start))
+            index += 1
+            continue
```
knisar, line 2013: [P2] Parse a fence on the list item's opening line

The unconditional `continue` skips the content after a list marker, so a valid opening fence there is never recognized. For example:

```python
output = (
    "Notices are required [adverse-action].\n\n"
    "- ```text\n"
    "  Rates [reg-z-2024]\n"
    "  ```"
)
result = _covering_scorer(output, CONTEXT).score(output, context=CONTEXT)
```

The reference renderer puts `Rates [reg-z-2024]` in a fenced code block inside the list, but extraction includes that marker. The scorer returns `score=0.0`, `passed=False`, and `fabricated_citations=['reg-z-2024']`, failing a response whose only actual citation is supported. Continue parsing the item's remaining content after consuming its marker, with offsets tied to the original response. Cover ordered and bullet markers, backtick and tilde fences, and LF/CR/CRLF; markers inside these blocks should stay excluded.

knisar (changes requested):
Reviewed 853c990 against current main at fe9c020. The previous nine list-paragraph regressions and six genuine-code controls now pass. The full existing suite passes with 648 passed and 3 skipped, and lint, REUSE licensing, patch whitespace, and all nine hosted checks are clean. The branch now merges cleanly, so I am removing the stale conflict label.

Two blocking correctness issues remain in the list-container scanner. First, a valid unindented continuation of a list paragraph closes the scanner's container too early, allowing a later fabricated citation in list prose to disappear and the row to pass. Second, a fenced code block beginning on the same line as a list marker is not opened, so a marker inside actual code is accused as a fabricated citation. The inline comments contain minimal reproductions and expected behavior.

I checked the block interpretations against the reference CommonMark parser. The first issue reproduces for ordered and bullet lists across LF, CR, and CRLF (6 failing assertions); the second reproduces for both list types, both fence characters, and all three line endings (12 failing assertions). These tests are additional to the passing existing suite. Please extend the container logic and regression corpus to cover both paragraph continuation and the first block on an item's opening line, while retaining the controls that genuine code stays excluded and list prose stays visible.

knisar:
Hi @M4h1m4, a rebase update: #69 and #70 have landed since my last review, and this branch now has a README conflict against current main (`5439934`). Please rebase and resolve that conflict alongside the two correctness findings in the review: the unindented list-paragraph continuation and a fence opening on the same line as a list marker.

When reconciling the README, retain the shipped scorer entries and documentation from main and add the citation-scorer entry. The `blocked: conflict` label now reflects this new conflict; the two existing review findings remain the substantive blockers.

knisar:
Hi @M4h1m4, checking in on the two remaining citation-scanner findings and the rebase. Please share a brief status and your expected next update by Friday, September 25.

The outstanding cases are an unindented continuation of a list paragraph and a fenced code block beginning on the same line as a list marker. The September 14 review has the reproductions and required controls. The branch still conflicts with current main, so please include the rebase with that revision.

If any part of the review is unclear or your availability has changed, please let me know. Thank you for the work through these edge cases.

PR #30, opened 2026-09-02
fix(scorers): preserve default LLM judge categories

### LINKED ISSUE #28: LLMJudgeScorer.__init__ wipes the class-level category when constructed directly
Surfaced while reviewing #27; the author flagged it there.

`BaseScorer.__init__` uses `None` as the sentinel for "keep the class attribute":

```python
if category is not None:
    self.category = category
```

`LLMJudgeScorer.__init__` defaults `category` to `""` and passes it through, so the sentinel never fires and every judge subclass loses its declared category when constructed directly:

```python
>>> FactualityJudge(api_key="x").category
''
>>> GroundednessScorer(api_key="x").category
''
```

All ten judge scorers are affected. It doesn't bite via the registry because `resolve_scorers` passes `category` explicitly, but anything constructed by hand (`additional_scorers`, notebooks, tests) ends up grouped under the scorer name in reports, and category-based policy expectations don't match.

Fix is a one-liner in `rai_toolkit/scorers/llm_judges.py`: change the default to `category: str | None = None`. Add a test asserting `FactualityJudge(api_key="x").category == "MIT-3.1"` and one for a subclass constructed with an explicit override.

YusefSyed:
## Summary

- let `BaseScorer` retain an LLM judge subclass's declared category when construction does not supply an override
- preserve both non-empty and explicit empty-string category overrides
- add focused regression coverage for all three constructor paths

Fixes #28.

## Why

`LLMJudgeScorer.__init__` previously defaulted `category` to `""` and always passed that value to `BaseScorer`. Because `BaseScorer` uses `None` to mean “keep the class attribute,” direct construction replaced categories such as `FactualityJudge.category == "MIT-3.1"` with an empty string.

Changing the default to `None` uses the existing sentinel contract. Callers that explicitly pass a category—including `category=""`—retain the previous override behavior.

## Validation

- `uv run --extra dev pytest tests/test_llm_judge_categories.py -q` — 3 passed
- `uv run --extra dev pytest tests/test_groundedness_scorer.py -q` — 6 passed
- `uv run --extra dev pytest -q` — 9 passed
- `uv run python -m compileall -q rai_toolkit tests` — passed
- `uvx --from 'reuse[charset-normalizer]' reuse lint` — REUSE 3.3 compliant

The existing test run reports 11 `asyncio.iscoroutinefunction` deprecation warnings from `rai_toolkit/_tracing.py`; this change does not introduce or modify them.

## AI assistance disclosure

AI tools assisted with repository search, implementation drafting, and an independent fresh-context review. I directed the work, reproduced the bug on the pinned upstream revision, reviewed the source and final two-file diff, added the compatibility edge case raised during review, and ran the validation listed above.

### DIFF
--- a/rai_toolkit/scorers/llm_judges.py
+++ b/rai_toolkit/scorers/llm_judges.py
@@ -96,7 +96,7 @@ class LLMJudgeScorer(BaseScorer):
         base_url: str | None = None,
         temperature: float = 0.1,
         threshold: float = 0.5,
-        category: str = "",
+        category: str | None = None,
         **kwargs: Any,
     ) -> None:
         super().__init__(category=category, threshold=threshold, **kwargs)
--- a/tests/test_llm_judge_categories.py
+++ b/tests/test_llm_judge_categories.py
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: 2026 Yusef Syed
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-PackageName: rai-toolkit
+
+from rai_toolkit.scorers import FactualityJudge
+
+
+def test_llm_judge_preserves_class_category_by_default() -> None:
+    scorer = FactualityJudge(api_key="test")
+
+    assert scorer.category == "MIT-3.1"
+
+
+def test_llm_judge_accepts_explicit_category_override() -> None:
+    scorer = FactualityJudge(api_key="test", category="custom")
+
+    assert scorer.category == "custom"
+
+
+def test_llm_judge_accepts_explicit_empty_category_override() -> None:
+    scorer = FactualityJudge(api_key="test", category="")
+
+    assert scorer.category == ""


knisar:
This is the right fix and the right size. Using the existing `None` sentinel rather than adding a new one keeps `category` consistent with how `BaseScorer` already handles `name`, `description` and `threshold`, and keeping the explicit `category=""` override working means nobody's existing call site changes meaning. Three tests for three constructor paths is exactly the coverage this needed.

Verified locally: your 3 tests pass, the PR head suite is green at 9, and merging this into current main after #24 gives 18 passing. It merges cleanly without a rebase.

Worth stating for anyone reading this later, since it is a behavior change even though it is the repair: any judge constructed directly without an explicit category now reports its real MIT category instead of an empty string. So pipeline summaries key on `MIT-3.1` rather than falling back to the class name, several direct judges in the same category now aggregate together, category-based policy triggers and composite weights start working, and trace names pick up the category suffix. No report field or type changes. Where same-category judges have different assessed-row counts, their now-correct co-aggregation can move an overall score and not just its label. That is what #28 was about.

Thanks for the clean writeup and for catching this. Squashing and merging. Fixes #28.

PR #31, opened 2026-09-02
fix: preserve additional scorers in Weave assessments

### LINKED ISSUE #29: Weave-native assessment path silently drops additional_scorers
Surfaced while reviewing #27, but it affects `GroundednessScorer` and any custom scorer today.

`Assessor._run_evaluation` copies `additional_scorers` onto the core pipeline, then, when `weave_project` is set and `weave` is installed, hands off to `WeaveEvaluationRunner.get_detailed_evaluation(...)` without forwarding them:

```python
if self.additional_scorers:
    self.pipeline.additional_scorers = list(self.additional_scorers)

if self._should_use_weave_evaluation():
    ...
    we_raw, we_summary = await runner.get_detailed_evaluation(
        weave_model,
        profile,
        weave_ds,
        name=ev_name,
        include_weave_builtins=self.include_weave_builtin_scorers,
    )
```

`get_detailed_evaluation` already accepts `additional_scorers: list[weave.Scorer] | None`, so the plumbing exists on the runner side; the assessor just never passes it. Result: with `weave_project=` set, every scorer supplied via `additional_scorers` is skipped with no warning, and the report shows no trace of it. Without `weave_project`, the same scorers run. That inconsistency is the bug.

This matters because `GroundednessScorer` (and `CitationCorrectnessScorer` in #27) are deliberately not registered in `scorer_registry.py`, so `additional_scorers` is the only way to run them. Right now they never run in a Weave-backed assessment.

Fix: wrap each extra `BaseScorer` with the existing Weave scorer adapter and pass the list through as `additional_scorers=`. If an extra scorer can't be adapted, log a warning rather than dropping it silently. Add a test that constructs an `Assessor` with `weave_project` set, a mocked runner, and one `additional_scorers` entry, and asserts the runner receives it.

YusefSyed:
Closes #29.

## Summary

- adapt caller-supplied toolkit scorers with the existing Weave scorer adapter
- forward the adapted scorers into `get_detailed_evaluation`
- preserve the documented `BaseScorer.score_async()` extension point in the Weave wrapper
- move inherited synchronous scorers and the bundled LLM judge bridge to worker threads so Weave retains row and scorer concurrency
- fall back to the core pipeline with every requested scorer if adaptation or Weave execution fails

## Verification

- the real offline Weave regression reproduces the pre-fix divergence: Weave called only the sync method and returned 0/fail while core called only the async method and returned 1/pass
- after the fix, the real Assessor, adapter, runner, Weave evaluation, and result converter use the Weave backend, call only the async method, preserve the exact scorer name and result cell, and match the core 1/pass result
- a 12-row real-Weave concurrency regression verifies overlapping calls for both the inherited base sync path and the LLM judge sync bridge while preserving every final result cell
- core environment without Weave: 20 passed, 3 optional tests skipped
- environment with Weave 0.53.7: 25 passed
- focused Ruff check and format check pass for the changed test module
- REUSE lint passes for all 106 files
- `git diff --check` passes for the scoped changes

## Base

The branch is rebased onto current `main` at `b3e81c3`, which includes merged #33 and its configured-scorer-name regression.

AI assistance was used to implement, test, and review this contribution.

### DIFF
--- a/integrations/weave_integration/scorers.py
+++ b/integrations/weave_integration/scorers.py
@@ -6,12 +6,14 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 from typing import Any
 
 import weave
 
 from rai_toolkit.scorers.base import BaseScorer, ScorerResult
+from rai_toolkit.scorers.llm_judges import LLMJudgeScorer
 
 logger = logging.getLogger(__name__)
 
@@ -63,6 +65,15 @@ def _filter_inner_scorer_kwargs(scorer: BaseScorer, extras: dict[str, Any]) -> d
     if any(p.kind == inspect.Parameter.VAR_KEYWORD for p in sig.parameters.values()):
         return dict(extras)
     return {k: v for k, v in extras.items() if k in sig.parameters}
+
+
+def _score_async_delegates_to_sync(scorer: BaseScorer) -> bool:
+    """Return whether score_async directly delegates to blocking score."""
+    implementation = type(scorer).score_async
+    return implementation in {
+        BaseScorer.score_async,
+        LLMJudgeScorer.score_async,
+    }
 
 
 class WeaveRAIScorer(weave.Scorer):
@@ -159,13 +170,24 @@ class WeaveRAIScorer(weave.Scorer):
         if plain_rubrics is not None:
             optional_inputs["rubrics"] = plain_rubrics
 
-        result = await self._rai_scorer.score_async(
-            output=output_text,
-            input=input_text,
-            context=effective_context,
-            **_filter_inner_scorer_kwargs(self._rai_scorer, optional_inputs),
-            **kwargs,
-        )
+        scorer_extras = _filter_inner_scorer_kwargs(self._rai_scorer, optional_inputs)
+        if _score_async_delegates_to_sync(self._rai_scorer):
+            result = await asyncio.to_thread(
+                self._rai_scorer.score,
+                output=output_text,
+                input=input_text,
+                context=effective_context,
+                **scorer_extras,
+                **kwargs,
+            )
+        else:
+            result = await self._rai_scorer.score_async(
+                output=output_text,
+                input=input_text,
+                context=effective_context,
+                **scorer_extras,
+                **kwargs,
+            )
 
         # Serialize manually instead of ``asdict`` so we can coerce
         # WeaveList/WeaveDict anywhere they sneak into ``details``
--- a/integrations/weave_integration/scorers.py
+++ b/integrations/weave_integration/scorers.py
@@ -109,7 +109,7 @@ class WeaveRAIScorer(weave.Scorer):
         call_display_name=lambda call: _wrapped_scorer_display_name(call),
         kind="scorer",
     )
-    def score(
+    async def score(
         self,
         output: dict[str, Any] | str,
         input_text: str = "",
@@ -159,7 +159,7 @@ class WeaveRAIScorer(weave.Scorer):
         if plain_rubrics is not None:
             optional_inputs["rubrics"] = plain_rubrics
 
-        result = self._rai_scorer.score(
+        result = await self._rai_scorer.score_async(
             output=output_text,
             input=input_text,
             context=effective_context,
--- a/rai_toolkit/assessment/assessor.py
+++ b/rai_toolkit/assessment/assessor.py
@@ -533,15 +533,7 @@ class Assessor:
 
                 weave_additional_scorers = []
                 for scorer in self.additional_scorers:
-                    try:
-                        weave_additional_scorers.append(make_weave_rai_scorer(scorer))
-                    except Exception as e:  # noqa: BLE001 - keep valid scorers running
-                        logger.warning(
-                            "Could not adapt additional scorer %s for Weave; "
-                            "skipping it (%s).",
-                            type(scorer).__name__,
-                            e,
-                        )
+                    weave_additional_scorers.append(make_weave_rai_scorer(scorer))
 
                 weave_ds = _dataset_rows_for_weave(dataset)
                 weave_model = WeaveModel(rai_model=self.model, model_name=self.model.name)
--- a/rai_toolkit/assessment/assessor.py
+++ b/rai_toolkit/assessment/assessor.py
@@ -520,11 +520,28 @@ class Assessor:
 
         if self._should_use_weave_evaluation():
             try:
-                from integrations.weave_integration.evaluation import WeaveEvaluationRunner
+                from integrations.weave_integration.evaluation import (
+                    WeaveEvaluationRunner,
+                )
+                from integrations.weave_integration.models import WeaveModel
+                from integrations.weave_integration.scorers import (
+                    make_weave_rai_scorer,
+                )
                 from rai_toolkit.evaluation.weave_adapter import (
                     weave_eval_results_to_evaluation_results,
                 )
-                from integrations.weave_integration.models import WeaveModel
+
+                weave_additional_scorers = []
+                for scorer in self.additional_scorers:
+                    try:
+                        weave_additional_scorers.append(make_weave_rai_scorer(scorer))
+                    except Exception as e:  # noqa: BLE001 - keep valid scorers running
+                        logger.warning(
+                            "Could not adapt additional scorer %s for Weave; "
+                            "skipping it (%s).",
+                            type(scorer).__name__,
+                            e,
+                        )
 
                 weave_ds = _dataset_rows_for_weave(dataset)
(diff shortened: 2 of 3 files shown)


knisar:
Good catch on the plumbing, and the call itself is right. I didn't want to take the mocks' word for it, so I installed the weave extra and checked against the real thing: `make_weave_rai_scorer` takes a plain `BaseScorer`, `additional_scorers` is the right keyword, `WeaveRAIScorer` does satisfy `weave.Scorer`, and a one-row eval runs through the real runner offline. 20 tests pass with and without the extra. The happy path closes #29.

I don't want to merge it as-is though, and it's the failure path that's stopping me.

The inner `except Exception` logs and carries on, so an assessment can come back green with a requested scorer quietly missing from it. Same inputs, two paths:

```
WARNING  Could not adapt additional scorer AdditionalScorer for Weave; skipping it.
evaluation_backend: weave
scorers_used: ['CoreScorer']
overall_score: 1.0    overall_passed: True
```

```
evaluation_backend: rai_pipeline
scorers_used: ['CoreScorer', 'AdditionalScorer']
overall_score: 0.5    overall_passed: False
```

Nothing in what comes back tells the caller coverage was reduced. That's the same problem #29 was about, just with a log line in front of it, and a warning in a log nobody reads doesn't really close it.

I'd let the failure escape the inner loop instead. Your outer handler already falls back to the core pipeline, and the additional scorers are installed there, so the fallback hands back a complete assessment rather than a partial one. That does mean `test_weave_evaluation_warns_and_keeps_adaptable_scorers` needs to change, since as written it pins the partial-loss behavior in place. Have it assert the core fallback.

Smaller thing on the tests. Faking the weave modules is the right call and keeps the extra out of the suite. But the fakes don't hold anything to the real contract: the stand-in adapter hands back a bare `object` and the fake runner swallows `**kwargs`, so these would still pass if someone renamed `additional_scorers` out from under you. Worth pinning the keyword and the element type, or mirroring the real signature in the fake.

One thing that is not yours, just so you know I saw it. `make_weave_rai_scorer` names wrappers off the class name rather than the configured `name`, so two instances of the same scorer class overwrite each other. Your PR is what makes that reachable, but it predates you and I'll open a separate issue for it. Nothing to do here.

Second PR from you today, on an issue that was four hours old when you picked it up. Thanks for the pace.

YusefSyed:
Thanks for the detailed check. I fixed the failure path in `3344713`.

An adapter error now reaches the existing outer Weave handler, so the assessment falls back to the core pipeline with every requested additional scorer. It no longer returns a partial Weave result.

I also tightened the fakes: the runner now has an explicit `additional_scorers` keyword, adapted scorers use a typed wrapper, and the fallback regression verifies that both requested scorers remain installed while the Weave runner is never reached.

The full suite passes at 20 tests, with focused Ruff and format checks, REUSE lint, and `git diff --check`. I left the separate configured-name collision out of this PR as suggested.

knisar (changes requested):
Thanks, I reran the final head at `3344713`. The fallback change does what I asked for: if the second scorer cannot be adapted, the Weave runner is never started and the core pipeline keeps and runs both original scorers. The full suite passes with and without Weave.

I found one more blocker by running an actual scorer through Weave instead of stopping at the mocked boundary.

The core pipeline calls `score_async()`, which is the documented extension point for async scorers. `WeaveRAIScorer.score()` calls the inner synchronous `score()` method instead. I tested a scorer whose sync path returns 0/fail and whose async path returns 1/pass. The Weave-backed assessment returned 0/fail with only the sync method called. The same assessment through the core pipeline returned 1/pass with only the async method called.

This also breaks a real bundled integration. `NeMoGuardrailScorer` works through the core path, but Weave runs its synchronous bridge in a worker thread where `asyncio.get_event_loop()` fails. That turns the same input into an assessed 0/fail solely because Weave is enabled.

Please preserve the async scorer contract before forwarding arbitrary `additional_scorers`. Either make the Weave wrapper await `score_async()`, or detect scorers with an async override and fall back to the core pipeline before starting the Weave run.

The current tests cannot catch this because they replace the Weave adapter, runner, and result converter, and never execute a scorer end to end. Please add an optional real offline Weave regression that checks the final scorer names and result cells, plus a core comparison for an async-overriding scorer.

There is also a merge-order dependency. I reproduced #32 with two same-class scorers named `alpha` and `beta`; Weave kept only one class-name result. I am not asking you to take over #33, but the corrected #33 should land before this PR.

Once the async path and end-to-end regression are in place, I can turn this around quickly.

YusefSyed:
I pushed the async-contract follow-up in `6e32876`.

- `WeaveRAIScorer.score()` is now async and awaits the documented `BaseScorer.score_async()` extension point instead of calling the synchronous method directly.
- The new optional real-Weave regression runs through the actual Assessor, adapter, runner, Weave evaluation, and result converter. It asserts that the result really used the Weave backend, checks the exact scorer name and result cell, and compares that cell with the core pipeline for an async-overriding scorer.
- Before the fix, the real Weave path called only the sync method and returned 0/fail while core called only the async method and returned 1/pass. After the fix, both paths call only the async method and return the same 1/pass cell.

Verification:

- core environment without Weave: 20 passed, 1 optional test skipped
- environment with Weave 0.53.7: 21 passed
- integration with the current #33 head `2ba73f`: 23 passed
- focused Ruff check and format check for the changed test module: passed
- REUSE: 105 of 105 files compliant
- `git diff --check`: passed for the scoped changes

I kept #33 separate as requested. The current #31 follow-up integrates cleanly with its head, and I will rebase after the corrected #33 lands.

knisar (changes requested):
Thanks for chasing the async mismatch. I reran `6e32876` against real Weave 0.52.40 and 0.53.7. The correctness side is fixed: the wrapper now reaches `score_async()`, the result matches the core pipeline, and the NeMo path no longer fails in a worker thread.

There is one new blocker before I can approve. Making the Weave op itself async moves every inherited sync scorer onto the event loop. `BaseScorer.score_async()` calls `score()` directly, and `LLMJudgeScorer.score_async()` does the same around blocking network I/O. Weave used to run the synchronous wrapper in a worker thread, so this change removes its row and scorer parallelism.

I reproduced it with twelve rows and a scorer that sleeps for 200 ms. The previous head completed in 0.99 seconds. `6e32876` took 3.24 seconds, with the same result on a repeat run. That is enough to turn normal assessments into long serial runs or timeouts.

Please keep awaiting real async overrides, but move the inherited synchronous path off the event loop, for example with `asyncio.to_thread()`. The bundled LLM judge fallback needs the same treatment. Please add a real-Weave concurrency regression so we hold onto both correctness and parallelism.

The original async blocker is resolved. #33 still needs to land first, then this branch should rebase before the final pass.

knisar:
#33 is merged now. Please rebase onto current main when you push the event-loop fix so the final run covers both pieces together.

YusefSyed:
I rebased onto current `main` at `b3e81c3`, which includes merged #33, and pushed the event-loop follow-up at `fd86d7e`.

The wrapper now distinguishes the two known synchronous bridges from true async overrides:

- scorers inheriting `BaseScorer.score_async()` run their synchronous `score()` method with `asyncio.to_thread()`
- bundled LLM judges inheriting `LLMJudgeScorer.score_async()` use the same worker-thread path
- genuine `score_async()` overrides are awaited directly

I added a 12-row real-Weave concurrency regression with both an inherited base sync scorer and an LLM judge sync bridge. It verifies the Weave backend, exact scorer names, every result cell, 12 calls per scorer, and overlapping calls greater than one for both scorers. The existing real-Weave async regression still verifies that true async overrides match the core pipeline and never call the sync method.

Verification on the rebased branch:

- core environment without Weave: 20 passed, 3 optional tests skipped
- environment with Weave 0.53.7: 25 passed
- focused Ruff check and format check for the changed test module: passed
- REUSE: 106 of 106 files compliant
- `git diff --check`: passed for the branch against current `main`

knisar (approved):
I reran the rebased head at `fd86d7e`. This clears the concurrency blocker.

On the same twelve-row probe that took 3.34 seconds on the previous head, this version finished in about 1.10 seconds. Both the inherited `BaseScorer` path and the bundled LLM judge bridge showed overlapping worker-thread calls, while a genuine async override still called only `score_async()` and matched the core result.

I also reran the real NeMo case, the complete adapter-failure fallback, and the two same-class scorers named `alpha` and `beta`. NeMo matches core, fallback retains every requested scorer, and both configured names and result cells survive now that #33 is in the base.

Final results: 20 passed with 3 optional skips without Weave, and 25 passed on both Weave 0.52.40 and 0.53.7. The Socket checks are green and the diff check is clean. Approved.

PR #33, opened 2026-09-02
fix(weave): use rai_scorer.name instead of class name in WeaveRAIScorer

### LINKED ISSUE #32: make_weave_rai_scorer names wrappers by class, so two instances of one scorer collide

```
rai_names: alpha beta
```

EffNine:
## Summary

`make_weave_rai_scorer` and `WeaveRAIScorer` name the wrapper from the scorer's **class** name rather than its configured `name`. Two instances of the same class with different names (e.g. two rubric scorers named 'alpha' and 'beta') write to the same Weave key and the later one silently overwrites the earlier:

```
rai_names: alpha beta
```

## Fix

Change `WeaveRAIScorer.__init__` to use `rai_scorer.name` (which already falls back to the class name via `BaseScorer.__init__`) instead of `type(rai_scorer).__name__`.

## Testing

Added `test_weave_rai_scorer_uses_configured_name_not_class_name` — two differently named instances of one scorer class both appear with their own names.
Added `test_weave_rai_scorer_falls_back_to_class_name_when_empty` — unnamed scorers still default to the class name.

All 20 tests pass.

Closes #32

### DIFF
--- a/integrations/weave_integration/scorers.py
+++ b/integrations/weave_integration/scorers.py
@@ -92,12 +92,12 @@ class WeaveRAIScorer(weave.Scorer):
     _rai_scorer: BaseScorer | None = None
 
     def __init__(self, rai_scorer: BaseScorer, **kwargs: Any) -> None:
-        # Set ``name`` (a weave.Scorer base field) to the rai scorer's class
-        # name so the Evals UI shows e.g. ``FairnessJudge`` in the scorer
-        # column instead of an unhelpful object-ref label. ``name`` also
-        # drives ``Scorer.display_name``.
-        rai_class = type(rai_scorer).__name__
-        kwargs.setdefault("name", rai_class)
+        # Set ``name`` (a weave.Scorer base field) to the rai scorer's
+        # configured ``name`` so that different instances of the same class
+        # (e.g. two rubric scorers with different names) remain distinct in
+        # the Evals UI. Falls back to the class name via BaseScorer when the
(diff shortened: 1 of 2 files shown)


knisar (changes requested):
Thanks for the focused fix. Using the configured scorer name is the right direction, and since this was the first linked implementation for #32, let's use #33 as the canonical PR.

Before merge, please tighten the regression coverage:

1. Add the real offline Weave evaluation case from #36, with two differently named instances of the same scorer class and assertions that both names and results survive in the final evaluation summary. The current wrapper-level assertions do not exercise the collision end to end.
2. Make that test optional-dependency aware. The core development environment does not install Weave, and this branch currently fails during test collection when Weave is absent. It should skip cleanly in that environment and run when the Weave extra is installed.
3. Confirm the SPDX copyright owner on the new test file and add the `SPDX-PackageName` header required by `CONTRIBUTING.md`.

Please keep the production change minimal. The stronger regression from #36 is the useful part to carry over.

knisar (approved):
I added the missing end-to-end coverage on top of the contributor's production fix and reran the branch in all supported setups. The no-Weave suite now skips cleanly, both configured names and result cells survive the real Weave evaluation and summary, and the SPDX header is complete.

Results: 18 passed and 1 skipped without Weave, 21 passed with Weave 0.52.40, and 21 passed with Weave 0.53.7. The production change remains focused. This is ready.

PR #34, opened 2026-09-02
fix(weave): use rai_scorer.name to prevent collisions between instances of same class (#32)

### LINKED ISSUE #32: make_weave_rai_scorer names wrappers by class, so two instances of one scorer collide

mhaye9545:
Closes #32.

## Problem

`make_weave_rai_scorer` in `integrations/weave_integration/scorers.py` previously named the Weave wrapper from the scorer's class name rather than its configured `name`.

When two instances of the same scorer class (e.g. `ConfigurableScorer(name="alpha")` and `ConfigurableScorer(name="beta")`) were included in an evaluation, both wrappers received `name = "ConfigurableScorer"`. Because Weave keys scorer outputs by `scorer.name`, the second instance overwrote the first in the evaluation summary without error.

## Solution

1. **Use configured scorer name**:
   - In `WeaveRAIScorer.__init__`, set `kwargs.setdefault("name", scorer_name)` where `scorer_name = rai_scorer.name or type(rai_scorer).__name__`.
   - In `make_weave_rai_scorer`, allow an optional `name` override, defaulting to `rai_scorer.name or type(rai_scorer).__name__`.
   - In `_wrapped_scorer_display_name`, format using `getattr(rai, "name", None) or type(rai).__name__`.
2. **Fallback**:
   - If `rai_scorer.name` is falsy, gracefully falls back to `type(rai_scorer).__name__`.

## Tests Added

- `tests/test_weave_scorers.py`:
  - `test_two_instances_of_same_scorer_class_have_distinct_names`: verifies two instances of one class have distinct `.name` attributes matching their configured names.
  - `test_empty_scorer_name_falls_back_to_class_name`: verifies fallback to class name when name is empty.
  - `test_custom_name_override_in_make_weave_rai_scorer`: verifies explicit name override in constructor helper.
  - `test_wrapped_scorer_display_name_uses_configured_name`: verifies call display name includes configured name.
  - `test_two_instances_in_weave_evaluation_do_not_collide`: verifies two differently named instances of one class both execute and produce distinct score entries without collision.

## Verification

- `pytest tests/test_weave_scorers.py` passes (5/5).
- `pytest tests/test_groundedness_scorer.py tests/test_hr_preset.py tests/test_llm_judge_categories.py` passes (18/18).
- `ruff check` passes.

### DIFF
--- a/integrations/weave_integration/scorers.py
+++ b/integrations/weave_integration/scorers.py
@@ -92,12 +92,8 @@ class WeaveRAIScorer(weave.Scorer):
     _rai_scorer: BaseScorer | None = None
 
     def __init__(self, rai_scorer: BaseScorer, **kwargs: Any) -> None:
-        # Set ``name`` (a weave.Scorer base field) to the rai scorer's class
-        # name so the Evals UI shows e.g. ``FairnessJudge`` in the scorer
-        # column instead of an unhelpful object-ref label. ``name`` also
-        # drives ``Scorer.display_name``.
-        rai_class = type(rai_scorer).__name__
-        kwargs.setdefault("name", rai_class)
+        scorer_name = rai_scorer.name or type(rai_scorer).__name__
+        kwargs.setdefault("name", scorer_name)
         super().__init__(
             rai_scorer_name=rai_scorer.name,
             rai_scorer_category=rai_scorer.category,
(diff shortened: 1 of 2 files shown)


knisar:
Thanks for the thorough alternative implementation. #33 was linked first and addresses the same bug with a smaller production change, so I am using #33 as the canonical PR for #32.

The useful additional coverage here, especially the configured display-name and override cases, can be carried into #33 if it remains relevant. Please close this as a duplicate so contributors are not maintaining two conflicting solutions to the same issue.

PR #36, opened 2026-09-02
fix: preserve configured names for Weave RAI scorers

### LINKED ISSUE #32: make_weave_rai_scorer names wrappers by class, so two instances of one scorer collide

YusefSyed:
Closes #32.

## Summary

- name each Weave wrapper from the configured RAI scorer name
- retain the scorer class name as a fallback
- prove that two differently named instances of one scorer class keep distinct result and summary keys

## Verification

- the real Weave evaluation collapsed both instances to one class-name key before the fix
- with the Weave extra: 19 passed
- without the Weave extra: 18 passed, 1 skipped
- focused Ruff and format checks passed
- REUSE lint passed
- `git diff --check` passed

AI assistance was used to implement and test this contribution.

### DIFF
--- a/integrations/weave_integration/scorers.py
+++ b/integrations/weave_integration/scorers.py
@@ -92,12 +92,12 @@ class WeaveRAIScorer(weave.Scorer):
     _rai_scorer: BaseScorer | None = None
 
     def __init__(self, rai_scorer: BaseScorer, **kwargs: Any) -> None:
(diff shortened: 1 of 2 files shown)


YusefSyed:
I missed that #33 and #34 were already linked before I claimed #32. They cover the same configured-name fix, so I am closing this duplicate to keep the review queue clean. This branch's real offline Weave regression remains available if it is useful to the maintainers.

PR #37, opened 2026-09-03
feat: add RetrievalRelevanceScorer for RAG retrieval quality evaluation

Zeming-Yuan:
> Re-submission of #25: GitHub auto-closed the original PR after my fork was accidentally deleted during an account cleanup on my side. The branch was recovered unchanged from the original head ref — same commits (`3f460a3`, `2de8a96`), so the two review fixes requested in #25 are already included.

Part of #4 (retrieval relevance track)

Adds `RetrievalRelevanceScorer`, an LLM-as-a-Judge scorer that evaluates
whether each retrieved context chunk is relevant to the user query. Designed
for RAG retrieval quality evaluation alongside the existing `GroundednessScorer`.

**Design:**
- Context contract: chunks are line-start `[source-id] text` blocks — the
  format the toolkit's reference RAG apps emit (see
  `demo_app/finance_advisor.py`) — with literal `---` delimiters kept as a
  fallback for unlabeled contexts. This matches the convention used by the
  citation scorers in #27.
- The judge reads the parsed chunks re-rendered into numbered
  `<chunk index="N">` envelopes, so a delimiter inside a chunk can never
  render as an extra boundary that a phantom verdict could grade.
- The judge assigns per-chunk relevance verdicts (`relevant`,
  `partially_relevant`, `irrelevant`); the overall 0-3 score is **derived
  from the validated verdicts** (a `partially_relevant` chunk weighs half),
  so the metric can never contradict them. The judge's own score is recorded
  in `details["judge_score"]` for audit.
- Refusal-shaped rows are still assessed: this scorer grades the retriever,
  and retrieved context stays gradable even when the generator declines.
- Verdict validation: dicts only, real-integer `chunk_index` (bools, floats,
  and strings are discarded, not coerced), unique and in-range, recognized
  label; any chunk without a verdict is `assessed=False` with
  `skipped="judge_parse_failure"` — a silent judge never reads as perfect
  retrieval.
- Judge-reply hardening: a valid-JSON reply with a non-object top level
  returns a controlled un-assessed result instead of crashing, and
  non-finite values (NaN/Infinity) in rejected replies are sanitized
  recursively so `details` stays strict-JSON serializable.

**Changes:**
- `rai_toolkit/prompts/judge_prompts.py`: prompt template for retrieval relevance
- `rai_toolkit/scorers/llm_judges.py`: `RetrievalRelevanceScorer` class
- `rai_toolkit/scorers/__init__.py`: export in import and `__all__`
- `README.md`: scorer list entry
- `tests/test_retrieval_relevance.py`: 32 tests covering all code paths

All 50 tests pass (18 existing + 32 new).


knisar (changes requested):
Thanks for restoring this after the fork deletion. I confirmed that #37 is byte-for-byte the same head as #25 (`2de8a96`), so the fixes from the earlier review are present. The full local suite passes with 35 tests, and the four-file diff is clean.

I found four integration and correctness issues that still need to be addressed before merge:

1. **The context parser does not accept the format produced by this repository's RAG apps.** The scorer splits chunks only on a literal `---`, while the finance and triage examples emit `[source-id] text` blocks separated by blank lines. With two native-format blocks, the scorer counted one chunk, discarded the second verdict, and still returned a passing score of 0.667. Please use a shared parser that understands the repository's source-labeled format, or accept structured chunks directly.

2. **The final score can contradict the validated verdicts.** A judge response with `score: 3` and two valid `irrelevant` verdicts currently returns `score=1.0`, `passed=True`, and `relevant_chunks=0/2`. Either derive the score from the validated labels or reject a payload whose overall score is inconsistent with them.

3. **`chunk_index` is coerced rather than validated.** `False`, `0.9`, and `"0"` are all accepted as index zero through `int(...)`. Require an actual integer and explicitly reject booleans, strings, and floats.

4. **The refusal skip conflicts with the retrieval-relevance contract in #4.** This metric grades the retriever, so retrieved context remains gradable even when the generator refuses. The implementation also checks `expected` rather than the output, and the broad `decline` marker causes factual text such as "Revenue declined 5 percent" to be skipped. Please remove the behavioral-refusal skip for this scorer and update the associated test.

A few cleanup items should land in the same update:

- #27 changes the same scorer file and uses the repository-native `[source-id]` context convention. Please align on one context contract and rebase after that shared behavior is settled.
- Change `Closes #4` to `Part of #4`, or close a dedicated child issue. #4 is the umbrella for several RAG and agent metrics.
- Update the PR description. This head contains 17 new test functions and 35 passing tests, not 7 and 13.
- Confirm the CoreWeave copyright assignment on the new external-authored test file, or use the contributor's correct copyright owner as required by `CONTRIBUTING.md`.
- Add regression cases for the native two-block context, contradictory scores, strict index types, and factual text containing "declined."

The dependency surface is unchanged and `git diff --check` passes. Once these points are fixed, I can re-review the focused four-file change quickly.

knisar (changes requested):
Thanks for the careful revision. I checked commit 94420c3 merged with current main. The 22 focused tests pass, the integrated core suite finishes at 42 passed and 3 skipped, and the configured Ruff, REUSE, and patch whitespace checks all pass. The native labelled format, derived scoring, strict index types, refusal behavior, issue wording, and copyright corrections all moved in the right direction.

I am keeping changes requested because four correctness cases remain:

1. The numeric result and displayed explanation can still disagree. With score 3, explanation "Perfect retrieval.", and two irrelevant verdicts, the result correctly returns score 0.0 and passed false but still says "Perfect retrieval." The displayed explanation should be derived from the validated verdicts. The original judge explanation can live in details for audit. Please extend the contradiction regression to assert this.

2. Duplicate indexes are still order-dependent. The current code keeps the first verdict and discards later ones. Reversing two conflicting index 0 verdicts changes an otherwise identical two-chunk result from a perfect pass to a failing one-third score, and both results remain assessed. A duplicate makes that chunk ambiguous. Treat duplicated indexes as a parse failure, or remove every verdict for a duplicated index so the missing-index check catches it. The current first-wins test needs to change.

3. Mixed context is parsed differently from what the judge receives. Once any label exists, _split_context_chunks() drops all text before the first label, while the full context still goes into the prompt. Both reference RAG apps create exactly this form when caller context is supplied before retrieved labelled blocks. This can shift judge indexes or exclude a noisy prefix from the derived score. Build the prompt from the exact parsed chunk sequence, or explicitly represent or reject the prefix, and add a regression for the reference-app form.

4. A Markdown link at the start of a line matches the source-label regex because no whitespace is required after the closing bracket. For a context such as "[docs](https://example.com) useful material" followed by a delimiter and an unrelated chunk, the parser counts one chunk instead of two and can return a perfect pass after discarding the second verdict. Require a source-label boundary that excludes Markdown-link syntax and add a regression.

These are all parser and result consistency fixes rather than new feature scope. Once they are addressed, the next pass should stay focused.

Zeming-Yuan:
Addressed all four correctness cases from your review (commit `9e55c75`):

1. No contradiction: the shown explanation is now derived from the validated verdicts (e.g. "0 relevant, 0 partially relevant, 2 irrelevant of 2 chunk(s); derived score 0/3..."); the judge's explanation from the review example is kept in `details["judge_explanation"]`. Extended in `test_contradictory_overall_score_is_derived_from_verdicts`.
2. Duplicated `chunk_index` is now a parse failure (`details["duplicate_chunk_indexes"]`), no longer first-wins - see `test_duplicate_chunk_indexes_are_a_parse_failure` (also with an off-scale second label).
3. The judge prompt is rebuilt from the exact parsed chunk sequence, so caller-context text before the first label is neither chunked nor shown to the judge - `test_caller_context_before_labelled_blocks_is_not_a_chunk`.
4. Source-label parsing now requires whitespace (or end of line) after the bracket, so `[docs](https://example.com)` at line start is no longer a label - `test_markdown_link_at_line_start_is_not_a_source_label`.

43 tests pass locally (22 focused + full suite), no new dependencies, `git diff --check` clean.

Zeming-Yuan:
Hardening follow-up (commit `80b6a0b`): the #27 sweep comment flagged the bool-subclasses-int class, and one instance was still here - the advisory judge score. `float(result.get("score"))` accepted `true` as 1.0. It is now rejected outright, the same typing rule the verdict chunk indexes follow: `details["judge_score"]` is null instead of a coerced bool. The metric itself is unchanged, since it is derived from the validated verdicts and never reads this field. Regression added (`test_boolean_judge_score_is_rejected_not_coerced`); 44 tests pass locally.

The parser is also aligned with `_parse_source_blocks` on #27: same character class and `(?=\s|$)` boundary, so Markdown links and reference definitions at line start are excluded on this side too.

knisar (changes requested):
Thanks for the follow-up. I checked both corrective commits through the exact head at 80b6a0b against current main. The four cases from my previous review are fixed: the explanation now follows the derived result, duplicate indexes fail deterministically, caller prefix text is absent from both the parsed count and prompt, and line-start Markdown links no longer become source labels. The latest boolean-score guard also behaves correctly.

The merge is clean. All 26 focused tests and all 92 repository tests pass locally, and the configured Ruff, REUSE, dependency, and whitespace checks pass. I still found two correctness cases that need changes.

1. A labelled chunk can contain the same delimiter used to render chunk boundaries. For context [fin-1] Revenue was $10M --- audited figure followed by [fin-2] Weather only, Python counts two labelled chunks, but the rebuilt prompt visibly contains three delimiter-shaped sections. A reasonable three-verdict reply for those visible sections is accepted: indexes 0 and 1 are treated as the two real chunks, index 2 is discarded, and the scorer returns assessed=true, score=1.0, passed=true even though the actual second source was the discarded irrelevant verdict. Please render explicit numbered chunk envelopes, or otherwise make the boundaries and indexes unambiguous without reusing content that may occur inside a chunk. Add a regression that inspects the prompt and exercises the extra-verdict mapping.

2. Valid JSON with a non-object top level is not handled. Replies of null, [], a string, or a number all raise AttributeError at result.get instead of returning a controlled un-assessed result. On the parse-failure path, a reply containing NaN, Infinity, or -Infinity is also retained in details.judge_response, so strict serialization with json.dumps(asdict(result), allow_nan=False) raises. Please guard for a JSON object before reading fields and sanitize rejected values recursively before storing them. The type and serialization contract now implemented on #27 is a useful reference, but this scorer needs its own regression coverage until the shared code is settled.

Please also refresh the stale test counts in the PR description and add RetrievalRelevanceScorer to the README scorer list. Since #27 changes the same parser area, this branch should be rebased after that shared contract lands so the helper resolution gets an explicit review.

Zeming-Yuan:
Both cases addressed (commit `e8e95d1`):

1. Chunks are now rendered as numbered `<chunk index="N">` envelopes instead of a `---` join, and the prompt reads exactly one envelope per index — a delimiter inside a chunk can no longer render as an extra boundary. The regression inspects the rendered prompt and exercises the extra-verdict mapping (the off-scale verdict for the phantom section is discarded, never graded).
2. A valid-JSON reply with a non-object top level (null, a list, a string, a number) returns a controlled un-assessed result, and non-finite values in rejected replies are sanitized recursively, so `details` stays strict-JSON serializable (`allow_nan=False`). Regressions cover all four non-object shapes and the NaN/Infinity case.

Also refreshed the PR description test counts (29 focused / 47 total), added the scorer to the README list, and noted the rebase after #27's shared contract lands.

knisar (changes requested):
Thanks for the focused update. I checked the exact head at `e8e95d1` merged with current main. Both findings from the previous review are fixed: delimiter text inside a labelled chunk no longer creates a delimiter-shaped section, and non-object JSON replies plus nested non-finite values now return controlled, strict-JSON-safe unassessed results. All 29 focused tests pass, the integrated suite finishes at 98 passed and 4 optional Weave skips, and Ruff, REUSE, dependency, and whitespace checks are clean.

I found two remaining boundary cases that need changes.

1. Retrieved text is inserted into the new XML-like envelopes without escaping. A chunk containing literal `</chunk>` and `<chunk index="1">` text can close the real envelope and forge another indexed envelope. In my two-real-chunk reproduction, the rendered prompt contained an injected index 1 envelope alongside the real index 1 envelope. A two-verdict reply was accepted as assessed with a 1.0 score even though the actual second source was irrelevant. Please make the representation collision-safe, either by escaping the envelope syntax inside chunk content or by using an unambiguous structured serialization, and add a regression with literal opening and closing envelope text in a retrieved chunk.

2. `_SOURCE_LABEL_PATTERN` uses `\s*` around the source ID. Since `\s` includes newlines, a context beginning with `[` followed by a newline and then `fin-1]` is accepted as a labelled block. That changes a two-section delimiter context into one labelled chunk, and one relevant verdict can produce an assessed full pass. Use horizontal whitespace such as `[ \t]*` around the ID and add a multiline-bracket regression.

The stated dependency on #27 also remains. I simulated #27 landing first, and merging this branch then conflicts in both `README.md` and `rai_toolkit/scorers/llm_judges.py`. After these two cases are fixed, please rebase once #27's shared parser contract lands and resolve that overlap explicitly before final review.

The new scorer is otherwise much closer. The derived scoring, duplicate-index handling, prompt alignment, non-object reply handling, and strict serialization all behaved correctly in this pass.

knisar:
One more validation case belongs in this correction pass. A verdict object whose `chunk_index` falls outside `range(len(chunks))` is currently counted as discarded, and the row can still remain assessed as long as every real index is present. I reproduced this with two real chunks plus a verdict for index 2: the result stayed assessed with a 1.0 score while the extra verdict was discarded.

Please treat any out-of-range integer index as `judge_parse_failure`, the same way duplicate indexes are handled, and record the offending indexes in `details`. Non-object verdicts and unrecognized labels can continue to be discarded because they do not claim a nonexistent chunk. Please update the forged-envelope regression to expect an unassessed result when the judge returns a phantom index.

Zeming-Yuan:
Thanks for the careful review. Both cases are fixed in two commits on the branch:

1. **Envelope injection** — chunk content is now XML-escaped (ampersand first) before rendering (`_escape_chunk_content`), so literal `</chunk>` or `<chunk index="1">` text inside a retrieved chunk can neither close the real envelope nor forge another one; the prompt contains exactly one envelope per chunk. Regression: `test_chunk_containing_envelope_text_is_collision_safe` (envelope openings equal the chunk count with literal opening and closing envelope text inside a chunk, and the full first chunk stays inside one envelope).

2. **Multiline bracket label** — the whitespace around the source ID is now horizontal only (`[ \t]*`): `\s` includes newlines, so `[\nfin-1]` read as a labelled block and collapsed a two-section delimiter context into one labelled chunk. The trailing lookahead still accepts a newline after the closing bracket. Regression: `test_bracket_before_newline_is_not_a_source_label` (the bracket-newline prefix no longer becomes a label; the context falls back to the delimiter split and both sections are chunks).

31 focused tests / 49 total pass locally; Ruff (E9,F63,F7,F82), REUSE, and `git diff --check` are clean. The PR description test counts are refreshed (31/49).

On #27: we will rebase once its shared parser contract lands and resolve the README.md / llm_judges.py overlap explicitly.

Zeming-Yuan:
Good catch -- that case is fixed too. A verdict whose integer `chunk_index` falls outside `range(len(chunks))` now fails the parse (`skipped="judge_parse_failure"`) exactly like a duplicate index, and the offending indexes are recorded in `details["out_of_range_chunk_indexes"]`. Type-level rejects (non-object verdicts, non-integer indexes) and unrecognized labels keep discarding, since they claim no chunk at all.

The forged-envelope regression now expects an un-assessed result when the judge returns a phantom index, and a new regression covers multiple offending indexes including negatives (`[0, 5, -1]` against two real chunks -> `out_of_range_chunk_indexes == [-1, 5]`, row un-assessed).

32 focused tests / 50 total pass locally; Ruff (E9,F63,F7,F82), REUSE, and `git diff --check` are clean. PR description test counts refreshed (32/50).

knisar (approved):
I rechecked e3c6509 after #54 landed. The final fixes are in place: retrieved chunk content is escaped before rendering, source-label whitespace no longer spans lines, and out-of-range integer indexes now fail parsing with audit details. The focused tests and integrated suite pass locally, and the branch merges cleanly with current main. This is ready to merge. #27 should rebase onto this version after it lands.

PR #46, opened 2026-09-04
fix(examples): preserve falsy values and validate limits

### LINKED ISSUE #40: fix(examples): preserve falsy dataset values and validate limits
## Problem

Generic row normalization uses truthiness when selecting aliases. Valid values such as numeric `0` and boolean `False` can be replaced with an empty string. Negative limits also behave differently between local and streaming loaders.

## Acceptance criteria

- Numeric `0` and boolean `False` values are preserved before string conversion.
- Missing values and `None` remain distinct from valid falsy values.
- An empty string may still fall through to a populated alias.
- `limit=0` returns no rows for every loader.
- Negative and boolean limits raise `ValueError`.
- Normalization does not mutate the original row.
- Existing rubric and policy expectation preservation remains unchanged.

## Non-goals

- A new dataset schema.
- Live Hugging Face tests.
- Changes to dataset content or licenses.

## Likely files

- `rai_toolkit/examples/registry.py`
- `tests/test_example_registry.py`

## Testing

Mock remote loaders. Cover local fixtures, generic rows, and at least one streaming loader.

## Claiming this issue

Comment `/claim` with the edge cases you plan to cover. Wait for `status: claimed` before starting. Keep the pull request offline and limited to loader contracts.

sansynx:
Closes #40

Preserves numeric zero and boolean false when normalizing dataset aliases, while missing, None, and empty values continue to fall through to populated aliases. Validates public load limits before dispatch so zero returns no rows and negative or boolean values raise ValueError.

Tests:
- python -m pytest -q
- python -m ruff check --select E9,F63,F7,F82 rai_toolkit/examples/registry.py tests/test_example_registry.py
- reuse --no-multiprocessing lint

AI-assisted contribution: I used an AI coding assistant during implementation and reviewed the complete diff and test results.

### DIFF
--- a/rai_toolkit/examples/registry.py
+++ b/rai_toolkit/examples/registry.py
@@ -131,7 +131,10 @@ class ExampleRegistry:
             and ``category``.
         """
         desc = ExampleRegistry.get(slug)
-        n = limit if limit is not None else desc.default_limit
+        n = _validate_limit(limit if limit is not None else desc.default_limit)
+
+        if n == 0:
+            return []
 
         if desc.loader is not None:
             rows = desc.loader(n)
@@ -154,10 +157,10 @@ def _normalize_row(row: dict[str, Any], default_category: str) -> dict[str, Any]
     assessable for a row.
     """
     out: dict[str, Any] = {
-        "input_text": str(row.get("input_text") or row.get("question") or row.get("prompt") or ""),
-        "context": str(row.get("context") or row.get("passage") or ""),
-        "expected": str(row.get("expected") or row.get("answer") or row.get("label") or ""),
-        "category": str(row.get("category") or default_category),
+        "input_text": str(_first_present(row, "input_text", "question", "prompt", default="")),
+        "context": str(_first_present(row, "context", "passage", default="")),
+        "expected": str(_first_present(row, "expected", "answer", "label", default="")),
+        "category": str(_first_present(row, "category", default=default_category)),
     }
     rubrics = row.get("rubrics")
     if rubrics:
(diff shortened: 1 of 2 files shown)


knisar (changes requested):
I checked 81494a7 against current main. The production change behaves correctly in the adversarial cases I ran: zero and false aliases survive normalization, missing and empty aliases fall through, the source row stays unchanged, and zero, negative, and boolean limits are rejected or short-circuited before all registered backends as intended. The focused run passes 5 tests, the full run passes 25 with 3 optional skips, and the configured Ruff, REUSE, dependency, and patch whitespace checks pass.

I am requesting changes for two missing regression cases from the issue contract:

1. The new loader tests only register an ExampleDescriptor with loader=.... They do not exercise the example_file branch, the generic huggingface_path branch, or one of the actual streaming loaders. Please add offline coverage for a temporary local JSON fixture, a generic Hugging Face descriptor with _require_datasets mocked, and at least one registered streaming loader with the same dependency mocked. Include a positive control showing each setup reaches its intended backend, then show zero and invalid limits do not invoke it. This is the behavior difference that originally varied between slicing and streaming.

2. Add False to the invalid-limit parameter set. False equals zero in Python, so an implementation that checked n == 0 before rejecting booleans would return an empty result instead of raising, while the current True-only boolean test would still pass.

The implementation itself does not need restructuring. This should be a focused test update.

sansynx:
Added the missing coverage for the file-backed, generic Hugging Face, and registered streaming paths, with positive controls plus zero, negative, and boolean limits. False is now covered as an invalid limit too

knisar (approved):
Thanks for filling in the missing regression coverage. I rechecked 78c3f6d against current main. The file-backed, generic Hugging Face, and registered TruthfulQA streaming paths now each have a positive control, and zero, negative, True, and False are checked before backend dispatch. Falsy alias preservation, fallback behavior, and non-mutation still hold.

Locally, the focused file passes 9 tests and the full suite passes 75. REUSE, dependency, and patch whitespace checks pass. The two broader Ruff findings in registry.py are pre-existing and outside this patch, while the configured CI lint selection is green. All hosted checks are green too.

This addresses my prior review. Approved.

PR #48, opened 2026-09-04
test(models): cover OpenAI-compatible adapter

### LINKED ISSUE #42: test(models): lock down the existing OpenAI-compatible endpoint contract
## Problem

The existing `OpenAICompatibleModel` already supports vLLM, Ollama, and similar local endpoints. The self-hosted guide was delivered in #5, and provider expansion is tracked in #21. What is still missing is a focused, offline contract test for this adapter.

## Scope

Test the current adapter behavior. This issue does not add another provider adapter.

## Acceptance criteria

- Mock `AsyncOpenAI` so the suite makes no network calls.
- Verify that the configured `base_url` and API key reach the client.
- Verify the non-empty fallback key used by local endpoints when no key is configured.
- Verify system, retrieved-context, and user message ordering.
- Verify default and per-call temperature behavior.
- Verify finish reason and token usage metadata.
- Verify `from_args()` handles its supported optional values and temperature conversion.

## Non-goals

- A dedicated Ollama or vLLM adapter.
- Starting a local inference server in tests.
- Anthropic, Gemini, or Bedrock support.
- A broad rewrite of the self-hosted guide.

## Likely files

- `rai_toolkit/models/openai_compatible.py`
- `tests/test_openai_compatible.py`

## Claiming this issue

Comment `/claim` with the fake client shape and cases you intend to test. Wait for `status: claimed` before starting. Keep production changes to small fixes exposed by the tests.

TheJhyeFactor:
## What changed

Adds a focused, fully offline contract suite for `OpenAICompatibleModel` and
`from_args()`.

The fake client records both its constructor arguments and each
`chat.completions.create(...)` call, then returns a deterministic response. The
tests cover:

- explicit base URL and API-key forwarding
- the non-empty fallback key used by local endpoints
- leaving an environment API key to the OpenAI SDK
- system, retrieved-context and user message ordering
- default and per-call temperatures
- output, finish reason and token-usage metadata
- `from_args()` defaults, empty optional values, explicit values and
  temperature conversion

No network connection, credentials, local model, or subprocess is used. The
adapter behaviour already matches the contract, so no production code needed
to change.

## Validation

- `python -m pytest -q tests/test_openai_compatible.py` — 8 passed
- `python -m pytest -q` — 28 passed, 3 skipped
- `python -m ruff check tests/test_openai_compatible.py` — passed
- `python -m ruff format --check tests/test_openai_compatible.py` — passed
- `python -m ruff check --select E9,F63,F7,F82 .` — passed
- `reuse --no-multiprocessing lint` — REUSE 3.3 compliant, 108/108 files
- `git diff --check` — passed

Tested with Python 3.12.14 on `darwin/arm64`.

## AI assistance

I used Codex to help inspect the adapter contract and draft the test harness. I
reviewed the fake-client behaviour, ran the full local suite and quality gates,
and can explain or rework any part of the change.

Closes #42

### DIFF
--- a/tests/test_openai_compatible.py
+++ b/tests/test_openai_compatible.py
@@ -1,0 +1,204 @@
+# SPDX-FileCopyrightText: 2026 Jhye
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-PackageName: rai-toolkit
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import Any, ClassVar
+
+import pytest
+
+from rai_toolkit.models import openai_compatible
+
+
+class FakeCompletions:
+    def __init__(self) -> None:
+        self.calls: list[dict[str, Any]] = []
+
+    async def create(self, **kwargs: Any) -> SimpleNamespace:
+        self.calls.append(kwargs)
+        return SimpleNamespace(
+            choices=[
+                SimpleNamespace(
+                    message=SimpleNamespace(content="offline answer"),
+                    finish_reason="stop",
+                )
+            ],
+            usage=SimpleNamespace(
+                prompt_tokens=7,
+                completion_tokens=3,
+                total_tokens=10,
+            ),
+        )
+
+
+class FakeAsyncOpenAI:
+    instances: ClassVar[list[FakeAsyncOpenAI]] = []
+
+    def __init__(self, **kwargs: Any) -> None:
+        self.kwargs = kwargs
+        self.completions = FakeCompletions()
+        self.chat = SimpleNamespace(completions=self.completions)
+        self.instances.append(self)
+
+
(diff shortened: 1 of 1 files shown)


knisar (approved):
I checked c98c2ca against current main and reproduced the focused and full suites with OpenAI SDK 2.2.0 and 3.8.0. I also exercised the adapter through an in-memory real-SDK transport to confirm the constructor arguments, three-message ordering, temperature override, and response metadata match the contract pinned by the fake.

The fixture restores the patched client and environment for every test, and the suite remains offline even with ambient OpenAI variables present. The focused run passes 8 tests. The full run passes 28 tests with 3 optional skips. Ruff, formatting, REUSE, dependency, and patch whitespace checks also pass. No changes requested.

PR #50, opened 2026-09-04
fix(scorers): reject non-finite normalized scores

### LINKED ISSUE #38: fix(scorers): reject non-finite normalized scores
## Problem

`ScoreNormalizer.from_scale()` currently allows non-finite values. In particular, `NaN` reaches the clamping logic and becomes `1.0`, so malformed judge output can turn into a passing score.

## Scope

Validate numeric inputs at the shared normalization boundary.

## Acceptance criteria

- Reject `NaN`, positive infinity, and negative infinity.
- Reject a non-finite `max_value`.
- Preserve the existing clamping and inversion behavior for finite values.
- A mocked judge response of `"NaN"` cannot produce an assessed pass through the core pipeline.

## Non-goals

- Redesigning every judge response schema.
- Changing finite out-of-range clamping.
- Changing citation parsing in #27 or retrieval parsing in #37.

## Likely files

- `rai_toolkit/scorers/normalizer.py`
- `tests/test_score_normalizer.py`

## Testing

Tests must be deterministic and make no API calls.

## Starting work

Before starting, check this issue's Development section and the open pull requests for linked work. Follow `CONTRIBUTING.md` by opening a draft pull request with `Closes #38`, your proposed validation behavior, and a short test plan. Keep the change limited to normalization and its regression tests.

hye654445:
Closes #38

## Problem

`min(1.0, float("nan"))` returns `1.0` in Python, so `ScoreNormalizer.from_scale()`
clamped a `NaN` raw score into a perfect score:

```python
>>> ScoreNormalizer.from_compliance_scale(float("nan"))
1.0
>>> ScoreNormalizer.apply_threshold(_, 0.7)
True
```

A judge returning `{"score": "NaN"}` therefore produced a passing compliance
result. `max_value` was unguarded for the same reason: `nan <= 0` is `False`,
so a non-finite scale slipped past the existing positivity check.

## Approach

**`normalizer.py`** — validate both arguments at the shared boundary. The finite
check on `max_value` deliberately runs *before* the positivity check, since a
`NaN` would otherwise pass the latter. Clamping and inversion are unchanged for
finite values.

**`llm_judges.py`** — at the two judge call sites, a non-finite raw score now
returns an un-assessed `ScorerResult` instead of propagating the `ValueError`.
`ScorerResult.assessed` already documents "parser failure" as its purpose, and
un-assessed results are excluded from aggregation, so one malformed row neither
inflates a score nor aborts an evaluation run.

### Question on scope

The issue lists only `normalizer.py` and its tests under "Likely files" and says
to keep the change limited to normalization. I touched `llm_judges.py` because
acceptance criterion 4 is about pipeline behaviour, and a bare `ValueError` there
would turn a silent wrong-pass into a crashed evaluation run.

Happy to split that half into a follow-up PR if you would rather keep this one
strictly at the normalization boundary — just say the word.

## Compatibility

`from_scale()` now raises `ValueError` for non-finite input where it previously
returned a clamped value. Any caller deliberately relying on `NaN → 1.0` or
`inf → 1.0` would be affected; within this repo the only callers are the two
judge sites, which are updated here.

## Tests

New `tests/test_score_normalizer.py` (22 cases):

- `NaN`, `+inf`, `-inf` rejected for `raw_score`, `max_value`, and inverted calls
- `from_compliance_scale` / `from_likert` reject `NaN`
- existing clamping, inversion, and the non-positive `max_value` error preserved
- end-to-end: a mocked judge returning `"NaN"` yields `assessed=False`,
  `passed=False`, `score=0.0`
- a finite judge response is still assessed and can still pass

I verified the tests actually catch the bug by reverting the source change and
re-running them: 14 of the 22 fail without the fix.

## Local validation

```
$ python -m pytest -q
42 passed, 3 skipped in 1.18s          # 20 passed, 3 skipped before this change

$ python -m ruff check --select E9,F63,F7,F82 .
All checks passed!

$ reuse --no-multiprocessing lint
Congratulations! Your project is compliant with version 3.3 of the REUSE Specification :-)
109 / 109 files with copyright and license information

$ git diff --check
(no output)
```

Python 3.11.4 on Windows.

## Unrelated observation

While setting up I noticed `rai_toolkit/redteam/attacks.py` is stored in the
repository with CRLF line endings, which contradicts `*.py text eol=lf` in
`.gitattributes`. A fresh `git clone` shows that file as modified immediately,
on a clean checkout with no local edits. It is untouched by this PR — I can open
a separate issue if that is useful.

---

AI-assisted: written with Claude Code. I have reviewed the change and can
explain and rework any line.

### DIFF
--- a/rai_toolkit/scorers/llm_judges.py
+++ b/rai_toolkit/scorers/llm_judges.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import json
 import logging
+import math
 from typing import Any
 
 from openai import OpenAI
@@ -179,6 +180,30 @@ class LLMJudgeScorer(BaseScorer):
         result = self._call_judge(prompts["system"], user_prompt)
 
         raw_score = float(result.get("score", 0))
+        if not math.isfinite(raw_score):
+            logger.warning(
+                "Judge %s returned a non-finite score (%s); marking un-assessed",
+                self.name,
+                raw_score,
+            )
+            return ScorerResult(
+                score=0.0,
+                passed=False,
+                category=self.category,
+                explanation=(
+                    "Un-assessed: the judge returned a non-finite score "
+                    f"({raw_score})."
+                ),
+                details={
+                    "skipped": "non_finite_judge_score",
+                    "scorer_name": self.name,
+                    "raw_score": raw_score,
+                    "max_score": 3,
+                    "judge_model": self.model,
+                    "judge_response": result,
+                },
+                assessed=False,
+            )
         normalized = ScoreNormalizer.from_compliance_scale(raw_score)
         passed = ScoreNormalizer.apply_threshold(normalized, self.threshold)
 
@@ -388,6 +413,31 @@ class GroundednessScorer(LLMJudgeScorer):
         user_prompt = self._format_prompt(output=output, input=input, context=context)
         result = self._call_judge(prompts["system"], user_prompt)
         raw_score = float(result.get("score", 0))
+        if not math.isfinite(raw_score):
+            logger.warning(
+                "Judge %s returned a non-finite score (%s); marking un-assessed",
+                self.name,
+                raw_score,
+            )
+            return ScorerResult(
+                score=0.0,
+                passed=False,
+                category=self.category,
+                explanation=(
+                    "Un-assessed: the judge returned a non-finite score "
+                    f"({raw_score})."
+                ),
+                details={
+                    "skipped": "non_finite_judge_score",
+                    "scorer_name": self.name,
+                    "raw_score": raw_score,
+                    "max_score": 3,
+                    "judge_model": self.model,
+                    "supporting_spans": [],
+                    "contradicting_spans": [],
+                },
+                assessed=False,
+            )
         normalized = ScoreNormalizer.from_compliance_scale(raw_score)
         raw_supporting = result.get("supporting_spans")
         raw_contradicting = result.get("contradicting_spans")
--- a/rai_toolkit/scorers/normalizer.py
+++ b/rai_toolkit/scorers/normalizer.py
@@ -6,6 +6,8 @@
 
 from __future__ import annotations
 
+import math
+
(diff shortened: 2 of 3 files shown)


knisar:
Keep the llm_judges.py handling in this pull request. The issue asks for a mocked NaN judge response to become unassessed through the core pipeline. Raising from the normalizer alone would replace the incorrect pass with a failed evaluation instead of meeting that contract. The extra file is justified as long as the change stays limited to handling non-finite scores at the existing judge boundaries.

Before you mark this ready, please tighten two details:

1. Do not place a float NaN or infinity in details["raw_score"]. That value can leak non-standard JSON into assessment reports. Store the original value as a string or another JSON-safe representation.

2. Teach _classify_unassessed_reason() to classify non_finite_judge_score as a scorer integration or parser failure, with a focused test. Otherwise the coverage report falls back to "see explanation in JSON report" instead of giving useful remediation.

No need to split this into another pull request. Please also rebase if #37 lands first, since both changes touch llm_judges.py.

knisar:
Hi @hye654445, just checking in on this one. This is still an important correctness fix, and `main` has moved since the draft was opened, so the branch now needs a rebase along with the two items in my comment above.

If you would still like to finish it, could you let me know by Friday, September 18? No worries at all if your availability has changed. If I do not hear back, I will plan to pick up the fix so it does not stay blocked.

Thank you again for getting this started.

knisar:
@hye654445, the September 18 response window has now ended without an update, so I am moving the active review for #38 to #68, as discussed. Thank you for starting this fix. I will keep this draft open until the replacement has been reviewed and accepted, then close it with a link to the landed implementation. Please join the discussion on #68 before resuming work here so we avoid duplicate effort.

knisar:
@hye654445, #68 has now passed final review and CI and merged, resolving #38. As promised after the September 18 response window and the handoff, I am closing this draft in favor of the landed implementation.

Thank you for the initial implementation and for getting this correctness fix started. The completed change is available in #68.

PR #51, opened 2026-09-04
fix(scorers): make CompositeScorer honor async and unassessed results

### LINKED ISSUE #39: fix(scorers): make CompositeScorer honor async and unassessed results
## Problem

`CompositeScorer` calls each child scorer's synchronous `score()` method. A child with a real `score_async()` override is therefore bypassed. The composite also lets unassessed child results affect aggregation, which conflicts with the `ScorerResult` contract.

## Acceptance criteria

- `CompositeScorer.score_async()` awaits each child's `score_async()` method.
- A real async override is exercised instead of its sync implementation.
- Unassessed child results do not affect weighted or unweighted scores.
- Weights are renormalized across assessed results.
- If every child is unassessed, the composite returns `assessed=False` with a useful reason.
- Result details include assessed and unassessed child counts.
- Existing synchronous callers continue to work.

## Non-goals

- Parallel child execution.
- Retry behavior.
- Changes to individual scorer semantics.

## Likely files

- `rai_toolkit/scorers/composite.py`
- `tests/test_composite_scorer.py`

## Testing

Use deterministic fake sync, async, failing, and unassessed scorers. No network access should be required.

## Starting work

Before starting, check this issue's Development section and the open pull requests for linked work. Follow `CONTRIBUTING.md` by opening a draft pull request with `Closes #39`, your intended aggregation behavior, and planned tests. Keep this separate from other scorer changes.

YusefSyed:
Closes #39.

`CompositeScorer` now honors genuine asynchronous child scorers and excludes unassessed results from aggregation, fail-fast decisions, and pass/fail counts. Weights renormalize over assessed children; a composite with no assessed children remains explicitly unassessed.

Child order stays sequential within each composite. Children inheriting the bundled sync-delegating `BaseScorer.score_async` or `LLMJudgeScorer.score_async` run through `asyncio.to_thread`, while genuine async overrides are awaited directly. This preserves concurrent Weave rows without changing the synchronous `score()` path.

Regression coverage includes weighted/unweighted and unassessed outcomes, mixed sync/async child ordering, execution-thread and context-variable preservation, exception propagation, and parallel composites through the real Weave evaluation adapter. The Weave regression fails on the earlier implementation because each synchronous child reaches only one active call; it passes after the change for both ordinary synchronous scorers and the bundled LLM-judge delegate.

Validation on current main:

- `WANDB_MODE=offline WANDB_SILENT=true .venv/bin/python -m pytest -q`: 66 passed with Weave 0.53.8.
- Focused composite/assessment tests with Weave 0.52.40: 12 passed.
- `.venv/bin/ruff check --select E9,F63,F7,F82 .`: passed.
- Ruff on all three touched files: passed.
- `.venv/bin/reuse --no-multiprocessing lint`: passed.
- `git diff --check`: passed.

AI assistance: AI tools assisted with investigation, implementation, test drafting, review, and validation. The account holder previously confirmed the contribution's ownership and Apache-2.0 submission rights and authorized this contribution and follow-up work. Tests used local fake scorers and offline Weave evaluation; no provider requests were made.

### DIFF
--- a/rai_toolkit/scorers/composite.py
+++ b/rai_toolkit/scorers/composite.py
@@ -66,44 +66,112 @@ class CompositeScorer(BaseScorer):
             result = scorer.score(output=output, input=input, context=context, **kwargs)
             results.append(result)
 
-        # Compute aggregate score
-        aggregate = ScoreNormalizer.aggregate_scores(results, self.weights)
+        return self._combine_results(results)
+
+    async def score_async(
+        self,
+        output: str,
+        input: str = "",
+        context: str = "",
+        **kwargs: Any,
+    ) -> ScorerResult:
+        """Asynchronously score each child before combining its result.
+
+        Child scorers run sequentially to preserve the composite's existing
+        execution behaviour; a child can override ``score_async`` for its own
+        asynchronous implementation.
+        """
+        results = []
+        for scorer in self.scorers:
+            result = await scorer.score_async(
+                output=output, input=input, context=context, **kwargs
+            )
+            results.append(result)
+
+        return self._combine_results(results)
+
+    def _combine_results(self, results: list[ScorerResult]) -> ScorerResult:
+        """Build a composite result from child results.
+
+        Un-assessed results are retained in details for coverage reporting but
+        never contribute to the aggregate or pass/fail decision.
+        """
+        assessed_results = [result for result in results if result.assessed]
+        unassessed_count = len(results) - len(assessed_results)
+
+        if not assessed_results:
+            return ScorerResult(
+                score=0.0,
+                passed=False,
+                category=self.category or "composite",
+                explanation="Un-assessed: no child scorers produced a signal.",
+                details=self._details(
+                    results,
+                    aggregate=0.0,
+                    assessed_count=0,
+                    unassessed_count=unassessed_count,
+                ),
+                assessed=False,
+            )
+
+        # Aggregate only assessed results, which also renormalizes weights.
+        aggregate = ScoreNormalizer.aggregate_scores(assessed_results, self.weights)
 
         # Determine pass/fail
         if self.fail_fast:
-            passed = all(r.passed for r in results)
+            passed = all(r.passed for r in assessed_results)
         else:
             passed = ScoreNormalizer.apply_threshold(aggregate, self.threshold)
 
         # Build explanation
-        failed_scorers = [r for r in results if not r.passed]
+        failed_scorers = [r for r in assessed_results if not r.passed]
         if failed_scorers:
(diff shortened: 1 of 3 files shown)


knisar (changes requested):
I checked 2010a69 against current main. The result semantics are otherwise solid: genuine async overrides are used, unassessed children are excluded from weighted and unweighted aggregation and fail-fast decisions, weights renormalize correctly, all-unassessed composites stay unassessed, and the new counts remain internally consistent. The focused run passes 5 tests, the full run passes 25 with 3 optional skips, and the configured quality checks pass.

One concurrency blocker remains. CompositeScorer.score_async() now directly awaits every child score_async(). Ordinary synchronous scorers inherit BaseScorer.score_async(), and the bundled LLM judges use another delegate with the same behavior. Both call the synchronous score() method inline. Because CompositeScorer now has its own async override, the Weave adapter treats it as genuinely asynchronous and no longer moves the whole composite into a worker thread.

I reproduced this with six concurrent composite evaluations containing one synchronous child that blocks for 100 ms. The new path took about 0.62 seconds, compared with about 0.11 seconds through the previous thread-offloaded path. That serializes Weave rows and can multiply evaluation time by the row count.

Please preserve sequential child order within one composite, but move sync-delegating children off the event loop while directly awaiting genuine async overrides. Add a concurrency regression for a composite containing a blocking synchronous child, ideally through the same adapter path that exposed the problem.

YusefSyed:
Addressed in 7c3ec18. Composite children retain sequential order, but the bundled BaseScorer/LLMJudge sync delegates now run off the event loop; genuine async overrides are still awaited directly. The real Weave evaluation regression now wraps both blocking scorer types in a composite and checks overlapping calls across rows. It fails before the fix and passes afterward. Additional tests cover mixed-child ordering, context propagation, execution threads, and exception handling. All 66 tests pass with Weave 0.53.8; the 12 focused composite/assessment tests also pass with 0.52.40. Ruff and REUSE checks pass.

knisar (approved):
Thanks, this addresses the concurrency blocker. I reran the real Weave regression against both revisions: it fails on the prior head because composite calls stay serialized, then passes on 7c3ec18 for both ordinary synchronous scorers and the LLM judge delegate. The test checks overlapping calls instead of relying on a timing cutoff.

Sequential order within a composite, true async dispatch, worker-thread routing, context propagation, nested composites, cancellation and exception behavior, and unassessed aggregation all checked out. The 12 focused tests and all 66 tests pass with both Weave 0.52.40 and 0.53.8. Configured Ruff, REUSE, dependency, and whitespace checks also pass. Approving this head.

PR #54, opened 2026-09-04
fix: use one toolkit version across packages and reports

Srijan229:
## Related issue

Closes #44

## What changed

* Added `rai_toolkit/_version.py` as the single version authority, retaining version `0.2.0`.
* Configured Hatchling to read package metadata from the shared version module.
* Updated `rai_toolkit.__version__` and `AssessmentResult.toolkit_version` to use the shared value.
* Preserved explicit historical version overrides in generated reports.
* Added regression coverage for package metadata, source checkouts, assessment results, JSON reports, and HTML reports.
* Updated the wheel smoke test to assert that the installed distribution version matches `rai_toolkit.__version__`.

## Validation

Local validation used Windows with Python 3.14:

* `python -m pytest -q tests/test_version.py` — 4 passed.
* `python -m pytest -q` — 64 passed, 4 skipped.
* `python -m ruff check --select E9,F63,F7,F82 .` — passed.
* `git diff --check` — passed.
* `python -m reuse --no-multiprocessing lint` — compliant, 114/114 files.
* `python -m build --no-isolation` — source distribution and wheel built successfully.
* Editable-install version smoke test — passed.
* Isolated wheel version smoke test — passed.
* Source-checkout test without distribution metadata — passed.
* `python -m pip check` — passed.

The Python 3.10–3.12 matrix and optional Weave environments were not reproduced locally and are left to CI.

## AI assistance

Codex assisted with repository inspection, implementation, regression tests, and local validation. I reviewed the complete diff and verified the reported local test results before submission.

## Checklist

* [x] I checked the issue's Development section and open pull requests for linked work.
* [x] This pull request is focused on one change.
* [x] Behavioural changes include tests, or I explained why a test is not needed.
* [x] I included the local validation commands and results.
* [x] Documentation is not required because this does not change user-facing behavior.
* [x] I checked ownership before adding or changing SPDX headers.

PR #56, opened 2026-09-07
feat(models): add Anthropic Messages API adapter

### LINKED ISSUE #55: Add an optional Anthropic model adapter
## Problem

The toolkit has a vendor-neutral `BaseModel` and an OpenAI-compatible adapter, but no native adapter for Anthropic's Messages API. This is the Anthropic provider slice from #21.

## Scope

Add an `AnthropicModel` backed by Anthropic's asynchronous Python client. Keep the SDK optional and preserve the existing `BaseModel` and `ModelResponse` contracts.

## Acceptance criteria

- Add `rai_toolkit/models/anthropic.py` with an `AnthropicModel` subclass and a `from_args()` helper.
- Support a required model name plus optional API key, base URL, system prompt, display name, and a positive `max_tokens` value with a documented default.
- Leave environment-based API key resolution to the Anthropic SDK when no key is supplied. Do not invent a fallback key.
- Send the system prompt and retrieved context through Anthropic's top-level `system` parameter. Send `input_text` as the user message.
- Allow `max_tokens` to be overridden per call. Pass only parameters supported by the current Anthropic Messages API.
- Return all text response blocks in `ModelResponse.output`.
- Normalize response metadata to the existing keys: `model`, `base_url`, `finish_reason`, `prompt_tokens`, `completion_tokens`, and `total_tokens`.
- Let `BaseModel.__init_subclass__` provide the standard `rai.model.predict` trace. Do not add a second tracing wrapper.
- Add an `anthropic` optional dependency extra and include it in `all`. Importing `rai_toolkit` and `rai_toolkit.models` must continue to work when that extra is not installed.
- Export `AnthropicModel` from `rai_toolkit.models` without making the optional SDK a core import-time dependency.
- Add fully mocked tests with no network calls. Cover client configuration, API key handling, system and context construction, default and per-call `max_tokens`, text extraction, metadata normalization, and `from_args()` conversion and validation.
- Document installation and a minimal usage example in the README, and add the dependency to the optional-dependency license table.

## Non-goals

- Adding Anthropic to the Streamlit intake form or changing `ApplicationProfile` routing.
- Streaming, tool use, prompt caching, multimodal content, or batch APIs.
- Anthropic through AWS Bedrock, Google Vertex AI, or Microsoft Foundry.
- A dedicated Ollama adapter. Ollama is already covered by `OpenAICompatibleModel`.
- Garak or PyRIT work. Those integrations already ship.
- Live API tests or tests that require credentials.

## Likely files

- `rai_toolkit/models/anthropic.py`
- `rai_toolkit/models/__init__.py`
- `tests/test_anthropic.py`
- `pyproject.toml`
- `README.md`

## Local validation

```bash
python -m pip install -e ".[dev,anthropic]"
python -m pytest -q tests/test_anthropic.py tests/test_openai_compatible.py
python -m pytest -q
```

## Starting work

Check the Development section for an existing linked pull request. Once you have an initial commit, open a draft pull request with `Closes #NN` in its description and list the checks you plan to run. Replace `#NN` with this issue number. The linked draft pull request is the signal that work has started, so no `/claim` comment is needed. Mark it ready only after the implementation and local validation are complete.

schallten:
## Related issue

Closes #55 

## What changed

Adds an optional `AnthropicModel` backed by Anthropic's async Messages API
client, exposed from `rai_toolkit.models` with a `from_args()` helper. The
anthropic SDK is imported lazily behind the new `[anthropic]` extra so the
core package keeps working without it.

System prompt and retrieved context go through the top-level `system`
parameter, output aggregates all text blocks, and response metadata is
normalized to the existing `ModelResponse` keys. Covered by fully mocked
offline tests for client configuration, API key handling, `max_tokens`
defaults and overrides, text extraction, metadata, and `from_args()`.

Behaviour / compatibility notes:

- `max_tokens` defaults to `1024`, is overridable per call, and rejects
  non-positive values.
- No env-var fallback key: resolution is left to the Anthropic SDK.
- Files touched: `rai_toolkit/models/anthropic.py` (new),
  `tests/test_anthropic.py` (new, 20 tests), `rai_toolkit/models/__init__.py`,
  `pyproject.toml`, `README.md`.

## Validation

- `pytest -q` (Python 3.10/3.11/3.12, `.[dev,anthropic]`): 89 passed, 4 skipped
- `pytest -q` (Python 3.12, `.[dev]` only): 69 passed, 5 skipped
- `pytest -q` (Python 3.12, `.[dev,weave]` + weave 0.52.40): 75 passed, 1 skipped
- `ruff check --select E9,F63,F7,F82 .`: clean
- `reuse --no-multiprocessing lint`: 115/115 compliant
- `python -m build`: wheel + sdist build; wheel installs in a clean venv,
  `pip check` clean, import works without the anthropic extra

## AI assistance

Implementation was AI-assisted (per CONTRIBUTING.md). SDK usage and local
validation were verified by the author; I can explain any line on request.

## Checklist

- [x] I checked the issue's Development section and open pull requests for
      linked work.
- [x] This pull request is focused on one change.
- [x] Behavioural changes include tests, or I explained why a test is not
      needed.
- [x] I included the local validation commands and results.
- [x] I updated documentation for user-visible changes.
- [x] I checked ownership before adding or changing SPDX headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,9 @@ pip install -e ".[pyrit]"
 # Garak red-team integration
 pip install -e ".[garak]"
 
+# Anthropic Messages API model adapter
+pip install -e ".[anthropic]"
+
 # Everything (incl. demo Streamlit app + dev tooling)
 pip install -e ".[all]"
 ```
@@ -304,6 +307,46 @@ git URL (e.g. `git+https://github.com/wandb/rai-toolkit.git@v0.1.0`).
 
 The Python import path is `rai_toolkit` (e.g. `from rai_toolkit import Assessor`)
 regardless of which install method you use.
+
+### Model adapters
+
+The toolkit ships vendor-neutral model adapters under `rai_toolkit.models`.
+Both expose a shared `BaseModel` / `ModelResponse` contract, so you can swap
+providers without changing your assessment code.
+
+```bash
+# Anthropic Messages API adapter (optional extra)
+pip install -e ".[anthropic]"
+```
+
+```python
+import asyncio
+
+from rai_toolkit.models import AnthropicModel
+
+
+async def main() -> None:
+    model = AnthropicModel(
+        model="claude-sonnet-4-5",
+        # No api_key? The Anthropic SDK falls back to ANTHROPIC_API_KEY.
+        system_prompt="You are a triage assistant.",
+        max_tokens=1024,  # positive default; override per call below
+    )
+    result = await model.predict(
+        "Review this answer.",
+        context="Retrieved policy text",
+        max_tokens=2048,
+    )
+    print(result.output)
+    print(result.metadata)
+
+
+asyncio.run(main())
+```
+
+The optional key is never substituted: when `api_key` is omitted, key
+resolution is left entirely to the Anthropic SDK (e.g. the
+`ANTHROPIC_API_KEY` environment variable).
 
 Running fully self-hosted or air-gapped? See [docs/self_hosted.md](docs/self_hosted.md).
 
@@ -465,7 +508,7 @@ rai_toolkit/
   assessment/         End-to-end Assessor workflow + HTML report
   workflow/              Review gate: profile, scoping, submission,
                          decisions, ManualFinding from interactive probing
-  models/                BaseModel + OpenAI-compatible adapter
+  models/                BaseModel + OpenAI-compatible + Anthropic adapters
   cli.py                 `rai` command
 
 integrations/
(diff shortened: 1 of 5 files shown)


knisar (changes requested):
Thanks for taking this on. I ran the focused adapter tests against both the older supported Anthropic SDK and the current SDK, and the full suite passes. Importing the core package without the optional dependency also works. I found three details that need correction before merge.

1. `max_tokens` is documented as a positive integer, but the constructor and per-call override accept `True` and `1.5`. `from_args()` also converts `True` to `1`, truncates `1.5` to `1`, and raises for blank or `None` values instead of using the documented default. Please validate a strict positive integer consistently at every entry point and add regressions for booleans, fractions, blank values, and `None`.

2. Multiple text response blocks are joined with `\n\n`, which inserts text that Anthropic did not return. Please preserve the block content without inventing separators and add a multi-block test that checks the exact output.

3. The module and `from_args()` docstrings say this path is used by Streamlit. #55 explicitly leaves Streamlit and `ApplicationProfile` routing out of scope, and this PR does not wire either path. Please describe `from_args()` as a configuration helper without claiming an integration that is not present.

Please also rebase onto current main now that #54 and #37 have landed. Once these points are addressed, the next pass should be focused.

schallten:
Fixed all three points (strict int max_tokens with None/blank defaulting, no separator between text blocks, Streamlit wording removed), rebased onto current main and validated.

knisar (approved):
I rechecked 3f073d4 against current main. All three requested fixes are in place, and the branch is clean. The focused adapter tests pass against both the declared Anthropic SDK floor and the current release, the full suite passes, and the optional import boundary, lint, and REUSE checks are clean. This is ready to merge.

PR #58, opened 2026-09-08
feat: add context precision and recall scorers for RAG retrieval

### LINKED ISSUE #26: Add context precision and recall scorers for RAG retrieval
Third slice of #4, after groundedness (#12, landed via #13) and citation correctness (#22).

Two scorers in `rai_toolkit/scorers/` that measure retrieval quality at the set level rather than chunk by chunk:

- **Context precision**: of the chunks that were retrieved, how many were actually needed to answer the query. Catches a retriever that pads the context window with noise.
- **Context recall**: of the information needed to answer the query, how much the retrieval actually supplied. Catches a retriever that misses the one chunk that mattered.

These pair with `RetrievalRelevanceScorer` (#37) but they are not the same measurement. Relevance grades each chunk on its own. Precision and recall grade the retrieved set as a whole, so a retriever can score well on relevance and still fail recall because of a chunk it never returned at all.

Conventions to follow, same as the existing judge scorers:

- Rows with no retrieved context return `assessed=False` rather than a defaulted zero.
- An empty or behavioural-refusal model response is not itself a reason to skip the row, because these scorers grade the retriever rather than the generator.
- Recall needs a reference answer or reference context on the row. Rows without one are un-assessed, not scored zero.
- Evidence spans follow the verified-span convention from #13, with the normalization added in #18.
- Tests mock `_call_judge`, so no key or network access is needed.

Implementation boundary after #37:

- Reuse the collision-safe source-label parsing and chunk conventions that are now on main. Do not introduce a second parser or an incompatible chunk format.
- Keep new `retrieved_ids` plumbing through `evaluation/pipeline.py` out of scope for this issue. Use the existing retrieved-context input for this slice. If structured ID propagation proves necessary, raise it in a separate focused issue rather than expanding this pull request.

Implementation is in progress in linked draft PR #58 from @Zeming-Yuan. The linked draft pull request is the work signal for this issue. No `/claim` comment or manual status change is needed.

Zeming-Yuan:
Closes #26.

Adds two set-level retrieval-quality scorers that grade the retrieved set as a whole, complementing `RetrievalRelevanceScorer` (#37), which grades each chunk on its own.

## What's added

- **`ContextPrecisionScorer`** — of the chunks that were retrieved, how many were actually needed to answer the query. A chunk that repeats information another retrieved chunk already supplies is `not_needed`, so padding and redundancy count against the score. The prompt's tie-break for identical chunks (lowest index is the needed one) is enforced by the scorer itself: chunks are grouped by passage content without the leading `[source-id]` label, so the same passage under two source IDs still groups as a duplicate, and within a group with any `needed` copy the lowest-index copy is the needed one — later copies marked `needed` are demoted and a lowest-index copy marked `not_needed` is promoted — with every correction recorded in `details["duplicate_chunk_corrections"]`.
- **`ContextRecallScorer`** — of the information needed to answer the query, how much the retrieval actually supplied. The judge decomposes the row's reference answer into individual pieces, each anchored to a verbatim span of the reference, and marks each with a JSON Boolean `supported` field plus a verbatim context span as evidence.

## Conventions followed

- Reuses the collision-safe chunk parsing (`_split_context_chunks`) and numbered envelope rendering (`_render_context_chunks`) that landed with #37; no second parser and no incompatible chunk format.
- Rows with no retrieved context return `assessed=False` (`skipped="empty_context"`) rather than a defaulted zero.
- Recall needs a reference: rows without `expected` return `assessed=False` with `skipped="missing_reference"`.
- Neither scorer skips refusal-shaped rows. These scorers grade the retriever, so retrieved context stays gradable when the generator declines to answer.
- Both derive their score from validated verdicts rather than the judge's own number; the judge's score is kept as an advisory value in `details["judge_score"]`.
- Recall anchors both sides: a piece whose reference span is not verbatim in `expected` is discarded, so a judge cannot invent information the reference does not contain; a `supported` piece whose context span is not verbatim is downgraded to not supported, so a judge cannot assert support it cannot show.
- Recall binds each piece to exactly one occurrence of its reference span: the earliest occurrence no earlier piece has claimed. A span the reference repeats must be listed once per occurrence, so one piece can never cover every repeated occurrence; a listing with no unclaimed occurrence left fails the parse rather than double-counting. The surviving pieces must also form a complete, non-overlapping decomposition of the reference (omitted pieces or overlapping spans fail the parse), so a judge cannot shrink the denominator.
- Recall scores the reference text the retrieval supplied, not judge items: adjacent pieces sharing a verdict are merged into maximal stretches of the reference, and the score is supported stretch length over total stretch length on the whitespace-collapsed reference — equivalent splits and merges leave the stretches unchanged, so the score does not depend on how the judge partitioned the same claims, while every additional unsupplied fact adds length the retrieval did not supply (`details["supported_span_length"]` / `details["total_span_length"]`).
- Context spans are verified against the original chunk content, not the XML-escaped envelope view: the judge's quote is unescaped, matched back to a retrieved chunk, and the stored span is the original chunk text, so a chunk containing `&` / `<` / `>` still verifies. The scaffolding check only strips real envelope markers and the matched chunk's own leading `[source-id]` label, so bracketed content such as `[FDA]` still verifies as evidence.
- Untrustworthy replies (non-object JSON, missing / duplicate / out-of-range verdicts, no verifiable reference piece, incomplete decomposition) return `assessed=False` with `skipped="judge_parse_failure"` instead of scoring.

## Out of scope

New `retrieved_ids` plumbing through `evaluation/pipeline.py` stays out of this PR, per the issue. If structured ID propagation proves necessary, it should be tracked in a separate focused issue.

## Tests

`tests/test_context_precision_recall.py` mocks `_call_judge`, so no key or network access is needed. On the branch: 164 passed, 4 skipped. Merged with current `main`: 208 passed, 4 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,7 +377,12 @@ present and retains verified supporting or contradicting spans as evidence.
 The `RetrievalRelevanceScorer` grades retrieval quality itself: it validates
 the judge's per-chunk relevance verdicts against the parsed chunks, derives
 the overall score from them, and returns un-assessed rather than silently
-passing when the judge reply cannot be trusted.
+passing when the judge reply cannot be trusted. The `ContextPrecisionScorer`
+and `ContextRecallScorer` measure the retrieved set as a whole rather than
+chunk by chunk: precision reports the fraction of retrieved chunks the query
+actually needed, and recall reports how much of the row's reference answer the
+retrieval supplied. Both derive their score from validated verdicts, so a
+judge that cannot be trusted leaves the row un-assessed instead of scoring it.
 
 ## Weave-native evaluation in assessment
 
--- a/rai_toolkit/prompts/judge_prompts.py
+++ b/rai_toolkit/prompts/judge_prompts.py
@@ -142,7 +142,9 @@ CONTEXT_PRECISION_SCORER_SYSTEM = (
     "a whole: a chunk that repeats information another retrieved chunk already "
     "supplies is not needed, even though it is on topic. Precision measures how "
     "much of the retrieved set was necessary, so redundant or off-topic chunks "
-    "count against it."
+    "count against it. When two chunks contain the same information, exactly one "
+    "of them is needed: the one with the lowest chunk index. Every later chunk "
+    "repeating that information is not needed."
 )
 
 CONTEXT_PRECISION_SCORER_TEMPLATE = """Decide whether each retrieved context chunk was needed to answer the user query.
@@ -158,7 +160,11 @@ Assign each chunk a label:
 
 Judge the retrieved set as a whole. A chunk that is on topic but repeats another
 chunk's information is "not_needed": precision asks how much of the retrieved set
-was necessary, not how much of it was on topic.
+was necessary, not how much of it was on topic. Apply one fixed rule when chunks
+repeat each other: if two chunks contain the same information, the chunk with the
+lowest chunk_index is the one that supplies it and is "needed"; every later chunk
+repeating that information is "not_needed", even when the chunks are otherwise
+identical.
 
 Score overall precision on a 0-3 scale:
 - 3: Every retrieved chunk was needed.
@@ -181,7 +187,9 @@ CONTEXT_RECALL_SCORER_SYSTEM = (
     "break a reference answer into the individual pieces of information it "
     "contains, then decide whether the retrieved context supplies each piece. "
     "Work only from the reference answer: every piece you list must be copied "
-    "verbatim from it, never paraphrased and never invented. A piece counts as "
+    "verbatim from it, never paraphrased and never invented. Decompose the "
+    "reference completely: the pieces must together cover the whole reference "
+    "answer, omit no part of it, and not overlap each other. A piece counts as "
     "supplied only when the retrieved context states it; topically related or "
     "partially matching content does not count."
 )
@@ -197,15 +205,21 @@ CONTEXT_RECALL_SCORER_TEMPLATE = """Decide how much of the reference answer the 
 Break the Reference Answer into its individual pieces of information: separate
 facts, figures, conditions, and recommendations. List each piece with an exact
 verbatim span copied from the Reference Answer. Do not paraphrase and do not add
-pieces the Reference Answer does not state.
+pieces the Reference Answer does not state. The pieces must together form a
+complete decomposition of the Reference Answer: every part of it belongs to
+exactly one piece, no part may be omitted, and no two pieces may overlap.
 
-For each piece, decide whether the Retrieved Context supplies it:
-- "supported": the retrieved context states this piece of information.
-- "not_supported": the retrieved context does not state it, or states only a
-  related but different fact.
+For each piece, decide whether the Retrieved Context supplies it, and record the
+verdict as a JSON Boolean in the "supported" field:
+- true: the retrieved context states this piece of information.
+- false: the retrieved context does not state it, or states only a related but
+  different fact.
+Use only the Boolean values true and false; never a quoted string label.
 
 For each supported piece, give an exact verbatim span copied from the Retrieved
-Context that states it. Spans must be copied exactly; do not paraphrase.
+Context that states it. Spans must be copied exactly; do not paraphrase. Quote
+the chunk's text itself, never the <chunk ...> envelope marker or the chunk's
+[source-id] label.
 
 Score overall recall on a 0-3 scale:
 - 3: Every piece of information in the reference answer is supplied by the retrieved context.
--- a/rai_toolkit/prompts/judge_prompts.py
+++ b/rai_toolkit/prompts/judge_prompts.py
@@ -134,6 +134,95 @@ Respond in JSON format:
   ]
 }}"""
 
+CONTEXT_PRECISION_SCORER_SYSTEM = (
+    "You are a strict retrieval-precision auditor for RAG systems. Your job is to "
+    "decide, for each retrieved context chunk, whether that chunk was actually "
+    "needed to answer the user query. A chunk is needed only if it supplies "
+    "information the query asks for. Judge each chunk against the retrieved set as "
+    "a whole: a chunk that repeats information another retrieved chunk already "
+    "supplies is not needed, even though it is on topic. Precision measures how "
+    "much of the retrieved set was necessary, so redundant or off-topic chunks "
+    "count against it."
+)
+
+CONTEXT_PRECISION_SCORER_TEMPLATE = """Decide whether each retrieved context chunk was needed to answer the user query.
+
+**User Query:** {input}
+
+**Retrieved Context (each retrieved chunk is wrapped in a numbered <chunk index="N">...</chunk> envelope):** {context}
+
+Return a per-chunk verdict. Each chunk is exactly one numbered <chunk index="N"> envelope above; use the envelope's index as the chunk_index.
+Assign each chunk a label:
+- "needed": the chunk supplies information the query asks for that no other retrieved chunk supplies.
+- "not_needed": the chunk is off-topic, or the information it supplies is already covered by another retrieved chunk.
+
+Judge the retrieved set as a whole. A chunk that is on topic but repeats another
+chunk's information is "not_needed": precision asks how much of the retrieved set
+was necessary, not how much of it was on topic.
+
+Score overall precision on a 0-3 scale:
+- 3: Every retrieved chunk was needed.
+- 2: Most chunks were needed, but at least one was redundant or off-topic.
+- 1: At most half of the chunks were needed.
+- 0: No chunk was needed.
+
+Respond in JSON format:
+{{
+  "score": <0-3>,
+  "explanation": "<brief reasoning about how much of the retrieved set was needed>",
+  "chunk_verdicts": [
+    {{"chunk_index": 0, "needed": "needed|not_needed", "reason": "<one short sentence>"}},
+    ...
+  ]
+}}"""
+
+CONTEXT_RECALL_SCORER_SYSTEM = (
+    "You are a strict retrieval-recall auditor for RAG systems. Your job is to "
+    "break a reference answer into the individual pieces of information it "
+    "contains, then decide whether the retrieved context supplies each piece. "
+    "Work only from the reference answer: every piece you list must be copied "
+    "verbatim from it, never paraphrased and never invented. A piece counts as "
+    "supplied only when the retrieved context states it; topically related or "
+    "partially matching content does not count."
+)
+
+CONTEXT_RECALL_SCORER_TEMPLATE = """Decide how much of the reference answer the retrieved context supplies.
+
+**User Query:** {input}
+
+**Retrieved Context (each retrieved chunk is wrapped in a numbered <chunk index="N">...</chunk> envelope):** {context}
+
+**Reference Answer:** {expected}
+
+Break the Reference Answer into its individual pieces of information: separate
+facts, figures, conditions, and recommendations. List each piece with an exact
+verbatim span copied from the Reference Answer. Do not paraphrase and do not add
+pieces the Reference Answer does not state.
+
+For each piece, decide whether the Retrieved Context supplies it:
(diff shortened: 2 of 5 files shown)


knisar (changes requested):
Thanks for putting this together. I ran the focused tests and the full suite, and all configured checks are green. The branch is cleanly based on current main. I found four correctness issues that need another pass before this is ready.

1. `ContextRecallScorer` derives its denominator only from the reference items the judge chooses to return. A judge can omit an unsupported fact from a multi-fact reference and the scorer still returns a perfect result. I reproduced a two-fact reference where only the first fact was retrieved. Returning only that supported item produced `assessed=True`, `score=1.0`, and `passed=True`. Overlapping or nested reference spans can also be counted as separate items. Please require a complete, non-overlapping reference decomposition or return `judge_parse_failure`, with regressions for omitted facts and overlapping spans.

2. Recall evidence is validated against the rendered XML envelope instead of the row's original context. A natural `R&D spending was $2M` quote is rejected, while `R&amp;D spending was $2M`, `<chunk index="0">`, and `[doc-1]` are accepted as support, and the escaped value is stored in the result. This breaks the verified-span convention from #13 and #18. Please validate only against original chunk content, map escaped prompt quotes back to original row offsets before storing them, and reject envelope or source-label-only spans. Please cover raw special characters, escaped prompt quotes, envelope-only spans, and label-only spans.

3. The recall prompt tells the judge to use `"supported"` and `"not_supported"` labels, but the JSON example and parser require a Boolean `supported` field. Please make the prose and schema consistently require `true` or `false`, and cover accepted Boolean values and rejected string values.

4. The precision prompt gives no deterministic rule for identical useful chunks. It says a repeated chunk is not needed because another chunk supplies the same information, which applies symmetrically to both copies. Please define an ordering rule, such as treating the earliest sufficient chunk as needed and later duplicates as not needed, and add an identical-chunk regression.

Once those points are addressed, the next review should be focused.

Zeming-Yuan:
Addressed all four cases (commit `e3cb3a8`):

1. The recall denominator now requires a complete, non-overlapping decomposition of the reference. After per-piece validation, the scorer maps every surviving reference span onto the normalized reference and fails the parse when spans overlap or any non-whitespace text is left uncovered (`skipped="judge_parse_failure"`, with `overlapping_reference_spans` / `uncovered_reference_text` recorded in details). The reproduced case -- a two-fact reference where only the supported fact is listed -- now returns un-assessed instead of 1.0. Regressions cover omitted facts and overlapping/nested spans.

2. Evidence is now verified against the original chunk content, never the rendered envelope. A quoted span is XML-unescaped (reversing `_escape_chunk_content`, ampersand resolved last, so a chunk that legitimately contained `&amp;` round-trips) and matched against the individual original chunks; the stored span is the original chunk text, not the escaped quote. A span consisting only of the envelope marker or the chunk's `[source-id]` label is rejected as evidence even when verbatim-present. Tests cover raw special characters, escaped quotes, envelope-only spans, and label-only spans.

3. The prose now requires the same Boolean the JSON example and parser already enforce, and both directions are covered by tests: `true`/`false` are accepted, string labels and other non-boolean values are discarded.

4. The precision prompt applies one deterministic rule when chunks repeat each other: the lowest-index chunk supplying a piece of information is "needed", every later chunk repeating it is "not_needed", even when the chunks are otherwise identical. An identical-chunk regression pins the score at 0.5 with the first chunk needed.

Ready for a focused re-review.

knisar (changes requested):
Thanks for the careful follow-up. I verified that the four earlier cases were addressed and tested the proposed merge with current `main`. The new scorer test file has 51 passing tests, the combined focused set has 144, and the full combined suite has 203 passed with 4 skipped. The configured Ruff check, REUSE check, and patch whitespace check also pass.

I found three remaining correctness cases before this is ready.

1. The recall denominator is still controlled by how the judge partitions the same reference. With `Alpha beta gamma. Delta epsilon.` as the reference and context supporting only the first sentence, two reference items produce `0.50`. Splitting only the supported sentence into `Alpha`, `beta`, and `gamma.` while keeping the same complete, non-overlapping coverage produces `0.75`. Both replies are accepted. `_normalized_occurrences()` also lets one item cover every repeated occurrence: `Apply twice. Apply twice.` with one supported `Apply twice.` item returns `1.0`. Please bind each item to one reference occurrence and make the derived recall invariant to equivalent split or merge choices. Add regressions for both cases.

2. The exact-duplicate precision rule is enforced only in the prompt. For context `same --- same`, a reply marking both chunks `needed` is accepted with `assessed=True` and `score=1.0`. Since the scorer can identify exact duplicate chunks itself, it should enforce the documented lowest-index rule or return a parse failure when the verdicts contradict it. Please add the inverse regression, where the judge marks both copies needed.

3. `_ENVELOPE_OR_LABEL_PATTERN` treats every bracket-only span as scaffolding. A legitimate reference and context value such as `[FDA]` is downgraded to unsupported even when it is the actual retrieved evidence. Please restrict this check to the real envelope markers and the actual leading source label for that chunk, then cover bracketed content that is not a source label.

Please also refresh the pull request description. It still mentions string `supported` / `not_supported` values and says evidence is checked against rendered envelopes, while the implementation now uses Booleans and original chunk text.

Once these are addressed, the next pass can stay focused on these regressions.

Zeming-Yuan:
Thanks for the careful follow-up — all three cases are real, and each now
has a fix plus a regression. The PR description is also refreshed (Boolean
`supported`, original-chunk-text verification, region-based recall, enforced
duplicate rule).

**1. Recall denominator controlled by the judge's partitioning** — two changes:

- *Occurrence binding.* Each item is bound to exactly one occurrence of its
  span: the earliest occurrence no earlier item has claimed. A repeated
  sentence must be listed once per occurrence — a single supported
  `Apply twice.` listing leaves the second occurrence uncovered (parse
  failure, `uncovered_reference_text`), and listing it twice scores 0.5.
  A listing with no unclaimed occurrence left is a duplicate parse failure.
  The decomposition check now consumes these bound intervals instead of
  re-deriving occurrences itself, so one item can never cover every repeat.
- *Partition invariance.* Adjacent pieces sharing a verdict are merged into
  maximal stretches of the reference before scoring, so the score follows
  which parts of the reference are supplied, not how the judge happened to
  split it. The two replies over `Alpha beta gamma. Delta epsilon.`
  (sentence-level vs. word-level split of the supported sentence) now both
  score 0.5. Details keys are `supported_regions` / `total_regions`.

**2. Exact-duplicate precision rule enforced only in the prompt** — the
scorer now groups chunks by whitespace-collapsed content (so `---` fallback
padding like `"same "` vs. `" same"` does not hide duplication) and demotes
any later exact duplicate the judge marked `needed`, recording each demotion
in `details["duplicate_chunk_corrections"]`. Your `same --- same` case:
both copies marked needed → 0.5, chunk 1 corrected.

**3. `_ENVELOPE_OR_LABEL_PATTERN` downgrades `[FDA]`** — the pattern now
matches only real `<chunk ...>` envelope markers, and label stripping is
restricted to the matched chunk's own leading `[source-id]` label
(`_context_span_in_chunks` returns the matched chunk for that purpose).
`[FDA]` in the chunk body verifies as evidence and is stored as support;
`[doc-1]`-label-only and envelope-only spans are still downgraded.

Regressions added: equivalent split invariance, repeated-sentence binding
(one listing → parse failure, two listings → 0.5), both-copies-needed
demotion, and bracketed content accepted as evidence. Scorer file: 56
tests; full suite: 161 passed / 4 skipped; ruff `--select E9,F63,F7,F82`
clean.

knisar (changes requested):
Thanks for the update. I rechecked head `8d7e1f4`. The three items from my previous review are resolved. The focused tests pass, and the proposed merge with current main is clean with 208 passed and 4 skipped.

Two scoring issues still block approval:

1. Duplicate detection includes the source label in its normalization. Identical passage text under different source IDs is therefore treated as distinct evidence. When both copies are marked needed, the scorer returns precision `1.0` with no correction. There is also an inverse case: for unlabelled duplicates, marking the first copy not needed and the second needed is corrected to both not needed, producing `0.0` instead of enforcing the documented lowest-index rule or rejecting the contradiction. Please normalize the passage content without its source label and cover both verdict directions.
2. Recall currently measures contiguous support runs rather than information supplied. I reproduced one supported fact followed by one, three, and ten distinct unsupported facts. Every case returned `0.5`; at the default threshold, one supplied fact out of eleven passes. This needs a metric decision before another narrow patch. The result should be independent of how equivalent facts are split while still decreasing as distinct unsupported information grows. Please define that behavior and add a regression that increases the number of missing facts.

Once these are addressed, I will rerun the focused and merged suites.

Zeming-Yuan:
Thanks for the third pass — both cases are real, and each now has a fix plus
a regression. The PR description is refreshed again (stretch-length recall,
label-stripped duplicate grouping, both-way lowest-index enforcement).

**1. Duplicate detection includes the source label** — two changes:

- *Label-stripped grouping.* Chunks are grouped by passage content without
  the leading `[source-id]` label, so identical passages under different
  source IDs now group as duplicates. Your case — the same passage under
  `[doc-1]` and `[doc-2]`, both marked needed — scores 0.5 with chunk 1
  corrected, instead of 1.0 with no correction.
- *Both verdict directions.* The lowest-index rule is now enforced
  symmetrically: within a group of identical chunks with any `needed` copy,
  the lowest-index copy is the needed one. A later copy marked `needed` is
  demoted as before, and a lowest-index copy marked `not_needed` while a
  later copy is needed is promoted, so your inverse case (`not_needed`,
  `needed`) resolves to (`needed`, `not_needed`) — 0.5 instead of the 0.0
  the one-sided demotion produced. A group no copy is needed in stays as
  judged, and every correction is recorded in
  `details["duplicate_chunk_corrections"]`.

**2. Recall measures contiguous runs, not information supplied** — the
metric is redefined: adjacent pieces sharing a verdict are merged into
maximal stretches, and the score is supported stretch length over total
stretch length on the whitespace-collapsed reference the pieces tile.

- Your reproduction: one supplied fact followed by one, three, and ten
  distinct unsupported facts now scores 24/41 ≈ 0.59, 24/77 ≈ 0.31, and
  24/204 ≈ 0.12 — strictly decreasing, where run counting returned 0.5 for
  all three.
- Partition invariance is kept by the merge, not lost by the length: an
  equivalent split or merge leaves the stretches unchanged, so the
  sentence-level and word-level decompositions of the same claims still
  score identically. Piece count alone would have moved with the partition
  (splitting one unsupported sentence into three pieces would triple the
  penalty), and run counting alone froze at 0.5 — weighing the stretches by
  length is what makes the score fall as missing information grows.
- Gaps between opposite-verdict stretches stay outside both, so the
  denominator is the stretches' own length, not the whole reference: the
  whitespace gaps the decomposition check allows depend on where the judge
  split spans, and only the stretches themselves are a partition-invariant
  measure. Details keys become `supported_span_length` /
  `total_span_length`.

Regressions added: the increasing missing-fact sequence (1 → 3 → 10 facts,
score must strictly decrease), source-label duplication under different
source IDs, and the contradictory lowest-index verdict in both directions.
Scorer file: 59 tests; full suite: 164 passed / 4 skipped; ruff
`--select E9,F63,F7,F82` and whitespace checks clean. Head is `551d959`.

knisar:
Thanks for taking another pass. I rechecked `551d959` against current `main`. The label-stripped duplicate grouping now works, and the lowest-index rule is enforced in both verdict directions. I also reproduced the growing-missing-content cases: recall falls as unsupported text grows while equivalent partitions remain stable. The proposed merge is clean, all current hosted checks are green, and the local focused and full checks pass. Please mark the PR ready when you are happy with the final description, and I will approve it.

knisar (approved):
Thanks for marking this ready. The head is still 551d959, which resolves the duplicate-source and recall-weighting issues from the previous review. I verified the proposed merge with current main at ec81e80: 335 tests passed and 3 skipped locally, and the syntax/undefined-name lint, REUSE licensing, and patch whitespace checks passed. All hosted checks are green. Approved.

PR #61, opened 2026-09-09
feat: add model adapter contract docs and conformance suite

abhnvgrg:
Closes #59

Adds the documented model adapter contract and a reusable, fully offline conformance test suite, exercised against both in-tree adapters.

What's included:

docs/model_adapters.md — adapter contract covering predict semantics, context handling, output fidelity, standard metadata keys, credential redaction, tracing, lazy imports, error propagation, and from_args() factories
tests/model_adapter_contract.py — reusable conformance suite (provider-SDK-agnostic, fully mocked)
Updated tests/test_openai_compatible.py and tests/test_anthropic.py to exercise the conformance suite while retaining all existing provider-specific tests
Updated CONTRIBUTING.md with a link to the adapter guide

Local validation:

python -m pytest -q tests/model_adapter_contract.py tests/test_openai_compatible.py tests/test_anthropic.py
83 passed, 2 skipped

python -m pytest -q
180 passed, 6 skipped (4 skips pre-existing)

python -m ruff check --select E9,F63,F7,F82 .
All checks passed!

reuse --no-multiprocessing lint
Congratulations! Your project is compliant with version 3.3 of the REUSE Specification :-)

git diff --check
(clean — CRLF auto-convert warnings only)

PR #64, opened 2026-09-10
feat(models): add a vendor-neutral CallableModel adapter

### LINKED ISSUE #60: Add a vendor-neutral CallableModel adapter for custom inference functions
## Problem

Teams should be able to evaluate an in-process model, internal SDK, or proprietary endpoint without adding a cloud-specific adapter or writing a new `BaseModel` subclass. A small callable wrapper would provide that vendor-neutral integration path without adding another provider dependency.

## Scope

Add a `CallableModel` that wraps a user-supplied synchronous or asynchronous inference function while preserving the existing `BaseModel` and `ModelResponse` contracts.

## Acceptance criteria

- Add `rai_toolkit/models/callable.py` with `CallableModel(BaseModel)` and export it from `rai_toolkit.models`.
- Use the public constructor `CallableModel(predict_fn, *, name: str | None = None)`. Reject a non-callable `predict_fn` immediately with `TypeError`. Pass `name` through normal `BaseModel` naming so omission falls back to `CallableModel`.
- Invoke the wrapped function as `predict_fn(input_text, context=context, **kwargs)`. Always forward `context`, including an empty string, and forward all call-time keyword arguments unchanged. Do not inspect or filter the callable's signature. Users can adapt a different signature with a lambda or `functools.partial`.
- Accept synchronous functions, asynchronous functions, partials, and callable objects. Await native asynchronous callables directly. Run synchronous callables with `asyncio.to_thread` so they do not block the event loop. If an invocation returns an awaitable, await it before normalizing the result.
- Accept only `str` and `ModelResponse` results. Convert a string to `ModelResponse(output=value)` with empty metadata. Return a `ModelResponse` unchanged, including its output and metadata.
- Raise `TypeError` for any other result and include the expected and actual result types in the message. Do not infer provider, token, latency, or model metadata. Callers that need metadata can return a `ModelResponse`.
- Propagate exceptions from the wrapped callable unchanged. Do not catch, log, retry, or convert them.
- Let `BaseModel.__init_subclass__` provide the single `rai.model.predict` trace. Do not import a tracing provider or add another tracing decorator.
- Add a short vendor-neutral README example showing both a simple string result and a metadata-bearing `ModelResponse`.
- Add no runtime dependency or optional extra.

## Tests

Add fully offline coverage in `tests/test_callable_model.py` for:

- A synchronous callable receiving input, context, and keyword arguments on a worker thread.
- An asynchronous callable receiving and returning the same values.
- A callable object or synchronous wrapper that returns an awaitable.
- String normalization to `ModelResponse` with empty metadata.
- A returned `ModelResponse` being preserved unchanged.
- Unsupported results including `None`, mappings, and bytes.
- Synchronous and asynchronous exceptions propagating without wrapping.
- Eager rejection of a non-callable constructor argument.
- Explicit and default names following `BaseModel` behavior.
- The inherited `rai.model.predict` tracing wrapper being applied exactly once.

## Non-goals

- Provider SDKs, credentials, or provider-specific request construction.
- Signature guessing or dropping unsupported keyword arguments.
- Dictionary or tuple response schemas.
- Streaming, batching, retries, timeouts, tools, or multimodal payloads.
- Cancelling a synchronous function that is already running on a worker thread.
- `from_args()`, CLI or `ApplicationProfile` wiring, serialization, entry-point registration, or remote callable transport.

## Likely files

- `rai_toolkit/models/callable.py`
- `rai_toolkit/models/__init__.py`
- `tests/test_callable_model.py`
- `README.md`

## Local validation

```bash
python -m pip install -e ".[dev]"
python -m pytest -q tests/test_callable_model.py tests/test_openai_compatible.py
python -m pytest -q
git diff --check
```

## Starting work

Check the Development section for an existing linked pull request. Once you have an initial commit, open a draft pull request with `Closes #NN` in its description and list the checks you plan to run. Replace `#NN` with this issue number. The linked draft pull request is the signal that work has started, so no `/claim` comment is needed. Mark it ready only after the implementation and local validation are complete.

M4h1m4:
Closes #60.

Wraps a user-supplied inference function so an in-process model, an internal SDK, or a proprietary endpoint can be assessed without writing a `BaseModel` subclass or adding a provider dependency.

```python
model = CallableModel(my_predict_fn, name="internal-sdk-v2")
result = await model.predict("Review this answer.", context="Policy text")
```

## Dispatch

Native coroutine functions are awaited directly. Everything else goes through `asyncio.to_thread`, so a synchronous function doing a network round trip does not hold the loop.

A synchronous callable can still return an awaitable, and `inspect.iscoroutinefunction` reports `False` both for a callable object with an `async __call__` and for a sync wrapper around an async client — so the dispatch cannot be decided from the function alone:

```python
result = await asyncio.to_thread(fn, input_text, context=context, **kwargs)
if inspect.isawaitable(result):
    result = await result
```

**One thing to confirm.** The criteria say to run synchronous callables with `to_thread` "so they do not block the event loop", and separately that an awaitable result is awaited before normalization, without fixing the sequence. Thread hop first is the reading that makes the first rule hold in the mixed case — awaiting on the loop directly would leave a blocking factory blocking it. Building a coroutine does not run it and coroutines are not thread-bound, so one created on the worker is safe to await on the loop. If you meant the other order, say so and I will switch.

## Everything else follows the criteria

The signature is never inspected, so a mismatch surfaces as Python's own `TypeError` and the caller adapts with a lambda or `partial`. A string becomes a response with empty metadata rather than an inferred one. A `ModelResponse` passes through untouched. Anything else raises `TypeError` naming both types. Exceptions propagate unchanged, a non-callable is rejected at construction, and `predict` is left for `BaseModel.__init_subclass__` so the trace stays single. `pyproject.toml` is untouched.

## Testing

30 offline tests covering the ten scenarios listed, plus three beyond them: that `context` is forwarded when empty, that a signature without `context` is left to raise rather than adapted, and that the blocking half of an awaitable factory runs off the loop.

Each guarantee was checked by removing it and confirming failures — dropping `to_thread` fails 2 tests, the awaitable await 3, the eager callable check 4, empty metadata 1, and raising on an unsupported result 8.

```
pytest -q tests/test_callable_model.py tests/test_openai_compatible.py   40 passed
pytest -q                                                               142 passed, 5 skipped
ruff check --select E9,F63,F7,F82 .                                     clean
reuse --no-multiprocessing lint                                         compliant
git diff --check                                                        clean
```

Branched directly on `6d0ed0c`.

## AI assistance

This contribution was AI-assisted (Claude). The design decisions and the dispatch reasoning above were worked through and reviewed by me, and I can explain or rework any line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,52 @@ Built-in adapters send `system_prompt` as trusted system instructions. When
 data instead of promoting it to the system channel. This preserves the role
 boundary, but does not by itself prevent prompt injection.
 
+### Evaluating your own inference function
+
+`CallableModel` wraps any function you already have, so an in-process model,
+an internal SDK, or a proprietary endpoint can be assessed without writing a
+`BaseModel` subclass or adding a provider dependency.
+
+```python
+import asyncio
+
+from rai_toolkit.models import CallableModel, ModelResponse
+
+
+def my_model(input_text: str, context: str = "", **kwargs) -> str:
+    """Return a plain string and the adapter wraps it."""
+    return internal_sdk.complete(input_text)
+
+
+def my_model_with_metadata(input_text: str, context: str = "", **kwargs):
+    """Return a ModelResponse to carry your own metadata through unchanged."""
+    reply = internal_sdk.complete(input_text)
+    return ModelResponse(
+        output=reply.text,
+        metadata={"model": reply.model, "total_tokens": reply.usage.total},
+    )
+
+
+async def main() -> None:
+    model = CallableModel(my_model, name="internal-sdk-v2")
+    print((await model.predict("Review this answer.", context="Policy text")).output)
+
+    detailed = CallableModel(my_model_with_metadata)
+    print((await detailed.predict("Review this answer.")).metadata)
+
+
+asyncio.run(main())
+```
+
+Synchronous functions run on a worker thread so they do not block the event
+loop; `async def` functions are awaited directly. The callable is always
+invoked as `predict_fn(input_text, context=context, **kwargs)` and its
+signature is never inspected, so adapt a differently shaped function yourself:
+
+```python
+model = CallableModel(lambda text, context="", **kwargs: legacy_predict(text))
+```
+
 Running fully self-hosted or air-gapped? See [docs/self_hosted.md](docs/self_hosted.md).
 
 ## Quickstart
--- a/rai_toolkit/models/__init__.py
+++ b/rai_toolkit/models/__init__.py
@@ -6,6 +6,13 @@
 
 from rai_toolkit.models.anthropic import AnthropicModel
 from rai_toolkit.models.base import BaseModel, ModelResponse
+from rai_toolkit.models.callable import CallableModel
 from rai_toolkit.models.openai_compatible import OpenAICompatibleModel
 
-__all__ = ["AnthropicModel", "BaseModel", "ModelResponse", "OpenAICompatibleModel"]
+__all__ = [
+    "AnthropicModel",
+    "BaseModel",
+    "CallableModel",
+    "ModelResponse",
+    "OpenAICompatibleModel",
+]
--- a/rai_toolkit/models/callable.py
+++ b/rai_toolkit/models/callable.py
@@ -1,0 +1,108 @@
+# SPDX-FileCopyrightText: 2026 M4h1m4
+# SPDX-FileCopyrightText: 2026 CoreWeave, Inc.
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-PackageName: rai-toolkit
+
+"""Vendor-neutral adapter wrapping a user-supplied inference function."""
+
+from __future__ import annotations
+
+import asyncio
+import inspect
+from collections.abc import Callable
+from typing import Any
+
+from rai_toolkit.models.base import BaseModel, ModelResponse
+
+__all__ = ["CallableModel"]
+
+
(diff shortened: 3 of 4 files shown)


knisar (changes requested):
The implementation looks correct, including the sync-first dispatch behavior. I ran the focused tests with 40 passing and the full suite with 182 passing and 4 skipped. Ruff, REUSE, and the diff check also pass.

One item needs clarification before approval. `rai_toolkit/models/callable.py` names CoreWeave, Inc. as a copyright holder. Please confirm that ownership of this contribution was assigned to CoreWeave. If it was not, remove that copyright line or replace it with the actual copyright holder while keeping the Apache-2.0 license identifier.

M4h1m4:
Done @knisar  - ownership was not assigned to CoreWeave. That line was mine in error and is removed, so the file now carries only my copyright with the Apache-2.0 identifier unchanged.

I have also rebased onto `e3186d6` now that #61 has merged, so the branch is two commits on current main with no conflicts.

162 tests pass on the rebased head, 30 of them on this adapter. Ruff, REUSE and the diff check are clean.

knisar (approved):
Thanks for fixing the copyright line and rebasing. I rechecked `5ff7323` against current `main`. The ownership concern is resolved, the adapter behavior still matches the contract, and the proposed merge is clean. I ran the focused tests and the full merged suite, and the licensing, lint, and whitespace checks all pass. Approved.

PR #65, opened 2026-09-11
fix(models): make adapter call-time options explicit

### LINKED ISSUE #63: Make built-in adapter call-time options explicit and reject unsupported kwargs
Child of #21. This follows #59 and should not be folded into PR #61, whose scope excludes public model API changes.

## Problem

`BaseModel.predict(..., **kwargs)` allows adapter-specific call options, but the built-in adapters do not say clearly which names they accept. `OpenAICompatibleModel` reads `temperature`, `AnthropicModel` reads `max_tokens`, and both silently discard every other key. A typo such as `max_token=512`, or code that swaps adapters while sending the same options, appears to work even though the provider request did not change.

## Scope

Make call-time options explicit for the two built-in provider adapters. A positive `max_tokens` value should work with both. `temperature` remains specific to `OpenAICompatibleModel`; it is not a portable option across current providers. Unsupported names must fail before any provider call instead of being ignored.

Implementation should wait for #61 to merge so the adapter guide and conformance suite can be updated in the same focused change.

## Acceptance criteria

- Keep `predict(input_text, context="", **kwargs)` on `BaseModel`. Do not add provider fields to the abstract signature.
- Continue supporting the existing per-call `temperature` override on `OpenAICompatibleModel`.
- Support a per-call `max_tokens` override on `OpenAICompatibleModel` and forward it to `chat.completions.create(...)`. When it is omitted, preserve the current request shape and provider default.
- Keep the existing `AnthropicModel` `max_tokens` default and per-call override behavior.
- Document a positive Python integer as the portable `max_tokens` form. `OpenAICompatibleModel` should accept only that direct-call form. Preserve `AnthropicModel`'s existing normalization for compatibility. Invalid values must fail before transport and the error must name `max_tokens`.
- In both built-in adapters, reject every unsupported call-time key with `TypeError` before constructing or invoking the provider client. The message must name the adapter and list all unsupported keys in deterministic order.
- Do not forward arbitrary keyword arguments to either provider SDK. In particular, callers cannot override adapter-controlled fields such as `model`, `messages`, `system`, or retrieved-context serialization.
- Add a short supported-options table to `docs/model_adapters.md` after #61 lands, and update both adapter docstrings. Make it clear that `max_tokens` is shared, `temperature` is OpenAI-compatible-specific, and unknown keys fail.
- Add fully mocked tests covering both accepted options, omitted-option request compatibility, a misspelled option, multiple unsupported options, deterministic errors, and confirmation that the transport recorder remains empty after validation fails.

## Compatibility

Existing documented calls keep working. `OpenAICompatibleModel.predict(..., max_tokens=N)` changes from a silent no-op to an effective limit. Calls that relied on ignored keyword arguments will now fail clearly, which is intentional and should be noted in the release notes.

## Non-goals

- Forwarding arbitrary SDK parameters.
- Making `temperature` a universal option.
- Standardizing `top_p`, stop sequences, seed, tools, streaming, retries, or multimodal input.
- Changing custom `BaseModel` subclasses.
- Changing `CallableModel`; #60 intentionally forwards arguments to the wrapped callable.
- Changing `GuardedModel` error handling or application routing.

## Local validation

```bash
python -m pip install -e ".[dev,anthropic]"
python -m pytest -q tests/test_openai_compatible.py tests/test_anthropic.py
python -m pytest -q
git diff --check
```

## Starting work

Wait for #61 to merge, then check this issue's Development section and the open pull requests. Open a draft pull request with `Closes #NN` once you have an initial commit. Replace `#NN` with this issue number.

The linked draft is the signal that work has started, so no `/claim` comment is needed.

dvd233:
Closes #63

## What changed

- Added a shared validation helper so built-in adapters reject unsupported call-time options with deterministic `TypeError` messages.
- Added the documented positive-integer `max_tokens` override to `OpenAICompatibleModel`, while preserving the provider-default request shape when it is omitted.
- Preserved `AnthropicModel`'s existing `max_tokens` normalization and made it explicit that `max_tokens` is the portable option; `temperature` remains OpenAI-compatible-specific.
- Kept adapter-controlled fields and arbitrary provider SDK parameters out of the call-time kwargs surface.
- Updated the adapter guide and added fully mocked regression coverage for accepted, omitted, invalid, misspelled, and multiple unsupported options.

The compatibility change is intentional: callers that relied on ignored keyword arguments now fail clearly instead of silently doing nothing.

## Validation

- `python -m pytest -q tests/test_openai_compatible.py tests/test_anthropic.py`: 104 passed, 2 skipped
- `python -m pytest -q` (proxy environment cleared): 206 passed, 6 skipped
- `python -m ruff check --select E9,F63,F7,F82 .`: passed
- `reuse --no-multiprocessing lint`: passed
- `git diff --check`: passed
- Changed-file Ruff, `compileall`, and `pip check`: passed

## AI assistance

AI assistance was used to inspect the scoped issue, draft the implementation and tests, and review edge cases. I personally reviewed the six-file diff and verified the behavior with mocked provider transports and the validation commands above.

## Checklist

- [x] I checked the issue's Development section and open pull requests for linked work.
- [x] This pull request is focused on one change.
- [x] Behavioural changes include tests, or I explained why a test is not
      needed.
- [x] I included the local validation commands and results.
- [x] I updated documentation for user-visible changes.
- [x] I checked ownership before adding or changing SPDX headers.

### DIFF
--- a/docs/model_adapters.md
+++ b/docs/model_adapters.md
@@ -47,9 +47,16 @@ With an empty `context` — the default — that means `input_text` is passed ve
 
 **An empty `context` must not create a synthetic context block.** `context=""` is the default and means "no retrieval happened". Emitting an empty `retrieved_context` field, or an empty "Retrieved context:" block, tells the model a retrieval returned nothing, which is a different claim from not having retrieved at all, and it changes behaviour.
 
-**`**kwargs` carries per-call overrides** of the adapter's own configured defaults. Today each built-in adapter reads exactly one key and silently ignores every other: `OpenAICompatibleModel` reads `temperature`, and `AnthropicModel` reads `max_tokens`. Unrecognized keys are dropped rather than forwarded to the provider, so a misspelled override fails silently — check the adapter you are calling before relying on one.
+**`**kwargs` carries per-call overrides** of the adapter's own configured defaults. The built-in adapters accept only the options in this table:
 
-An adapter that does read a key validates it with the same rule its constructor uses: `AnthropicModel` runs a per-call `max_tokens` through `_coerce_max_tokens`, so an invalid value raises `ValueError` before the request is built. A general `**kwargs` API — an agreed set of overrides and rejection of unknown or invalid keys — is follow-up work tracked in [#63](https://github.com/wandb/rai-toolkit/issues/63). A new adapter should not invent its own until that lands.
+| Adapter | Supported call-time options | Validation and request behaviour |
+| --- | --- | --- |
+| `OpenAICompatibleModel` | `temperature`, `max_tokens` | `max_tokens` must be a positive Python integer. It is omitted from the provider request when the caller does not supply it, preserving the provider default. |
+| `AnthropicModel` | `max_tokens` | A positive Python integer is the portable form. The adapter also preserves its existing normalization of integer strings, `None`, and blank strings for compatibility. |
+
+Every other call-time option raises `TypeError` before a provider request is made. The error names the adapter and lists every unsupported key in sorted order, so misspellings such as `max_token` cannot look successful while doing nothing. Adapter-controlled fields such as `model`, `messages`, `system`, and retrieved-context serialization cannot be overridden through `**kwargs`.
+
+This is an intentional compatibility change for callers that relied on ignored keyword arguments: those calls now fail clearly. Provider SDK parameters are not forwarded automatically; adding another supported option requires an explicit adapter contract, validation, documentation, and tests.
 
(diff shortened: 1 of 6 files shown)


knisar (approved):
Thanks for taking this on. I checked the adapter behavior, error handling, documentation, and mocked coverage at 9ffc4f3. I also ran the full suite and tested the result against current main. The two Weave jobs used the PR's older base and hit the test package import issue that is already fixed on main. The merged result passes locally. This is ready to merge.

PR #68, opened 2026-09-11
fix(scorers): reject non-finite normalized scores

### LINKED ISSUE #38: fix(scorers): reject non-finite normalized scores

abhnvgrg:
Closes https://github.com/wandb/rai-toolkit/issues/38

What
Non-finite values (NaN, +inf, -inf) can enter the scoring pipeline
from malformed LLM judge responses. These values poison score aggregations
and can silently pass or fail policy gates.

Fix
ScorerResult.__post_init__: explicit math.isfinite() guard alongside
the range check, with a clearer error message
ScoreNormalizer.from_scale: reject non-finite raw_score and max_value
before normalization
ScoreNormalizer.aggregate_scores: filter to assessed=True results only,
preventing un-assessed placeholders from diluting real scores
LLMJudgeScorer.score: catch non-finite raw judge scores and return
assessed=False rather than propagating the bad value
Parametrized tests for NaN, +inf, -inf at each validation layer
Regression guards confirming valid scores still work
AI disclosure
Implementation assisted by Claude. All code reviewed and tested locally.

Local validation
pytest tests/test_nonfinite_scores.py -v
17 passed

pytest tests/ -x -q
343 passed, 14 skipped

### DIFF
--- a/tests/test_nonfinite_scores.py
+++ b/tests/test_nonfinite_scores.py
@@ -1,0 +1,86 @@
+# SPDX-FileCopyrightText: 2026 Abhinav Garg
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-PackageName: rai-toolkit
+
+"""Non-finite scores must be rejected at every layer, not silently averaged in."""
+
+from __future__ import annotations
+
+import pytest
+
+from rai_toolkit.scorers.base import ScorerResult
+from rai_toolkit.scorers.normalizer import ScoreNormalizer
+
+
+@pytest.mark.parametrize("bad_value", [float("nan"), float("inf"), float("-inf")])
+def test_scorer_result_rejects_nonfinite(bad_value: float) -> None:
+    with pytest.raises(ValueError, match="finite"):
+        ScorerResult(score=bad_value, passed=False, category="test", explanation="x")
+
+
+@pytest.mark.parametrize("bad_value", [float("nan"), float("inf"), float("-inf")])
+def test_normalizer_from_scale_rejects_nonfinite_raw_score(bad_value: float) -> None:
+    with pytest.raises(ValueError, match="finite"):
+        ScoreNormalizer.from_scale(bad_value, max_value=3.0)
+
+
+@pytest.mark.parametrize("bad_value", [float("nan"), float("inf"), float("-inf")])
+def test_normalizer_from_scale_rejects_nonfinite_max_value(bad_value: float) -> None:
+    with pytest.raises(ValueError):
+        ScoreNormalizer.from_scale(1.0, max_value=bad_value)
+
+
+def test_aggregate_scores_skips_unassessed_results() -> None:
+    results = [
+        ScorerResult(
+            score=0.8, passed=True, category="a", explanation="ok", assessed=True
+        ),
+        ScorerResult(
+            score=0.6, passed=True, category="b", explanation="ok", assessed=True
+        ),
+        ScorerResult(
+            score=0.0, passed=False, category="c", explanation="skip", assessed=False
+        ),
+    ]
+    avg = ScoreNormalizer.aggregate_scores(results)
+    # Should average only the two assessed results: (0.8 + 0.6) / 2 = 0.7
+    assert avg == pytest.approx(0.7)
+
+
+def test_aggregate_scores_weighted_skips_unassessed_results() -> None:
+    results = [
+        ScorerResult(
+            score=0.8, passed=True, category="a", explanation="ok", assessed=True
+        ),
+        ScorerResult(
+            score=0.0, passed=False, category="b", explanation="skip", assessed=False
+        ),
+    ]
+    avg = ScoreNormalizer.aggregate_scores(results, weights={"a": 2.0, "b": 5.0})
+    assert avg == pytest.approx(0.8)
+
+
+def test_aggregate_scores_all_unassessed_returns_zero() -> None:
+    results = [
+        ScorerResult(
+            score=0.0, passed=False, category="a", explanation="skip", assessed=False
+        ),
+    ]
+    assert ScoreNormalizer.aggregate_scores(results) == 0.0
+
+
+def test_aggregate_scores_empty_returns_zero() -> None:
+    assert ScoreNormalizer.aggregate_scores([]) == 0.0
+
+
+# Regression guards -- valid values still work
+@pytest.mark.parametrize("good_value", [0.0, 0.5, 1.0])
+def test_scorer_result_accepts_valid_finite_scores(good_value: float) -> None:
+    result = ScorerResult(
+        score=good_value, passed=True, category="test", explanation="ok"
+    )
+    assert result.score == good_value
+
+
+def test_normalizer_from_scale_valid() -> None:
+    assert ScoreNormalizer.from_scale(2.0, max_value=3.0) == pytest.approx(2.0 / 3.0)
--- a/rai_toolkit/scorers/base.py
+++ b/rai_toolkit/scorers/base.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+import math
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from typing import Any
@@ -36,8 +37,10 @@ class ScorerResult:
     assessed: bool = True
 
     def __post_init__(self) -> None:
-        if not 0.0 <= self.score <= 1.0:
-            raise ValueError(f"Score must be between 0.0 and 1.0, got {self.score}")
+        if not math.isfinite(self.score) or not (0.0 <= self.score <= 1.0):
+            raise ValueError(
+                f"Score must be a finite number between 0.0 and 1.0, got {self.score}"
+            )
 
 
 class BaseScorer(ABC):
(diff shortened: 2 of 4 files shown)


knisar:
Hi @abhnvgrg, thanks for putting this together and for including focused tests.

PR #50 is currently the active implementation for #38, and its contributor work window runs through Friday, September 18. To avoid having two contributors solve the same issue at the same time, I am marking this PR blocked while work on #50 remains active.

Please pause additional work here for now. If #50 is not continued, I will come back to this PR as the next candidate implementation. No changes are needed from you while it is blocked. Thanks for understanding, and I appreciate the contribution.

abhnvgrg:
Thanks for the heads-up, @knisar. I totally understand. Happy to wait and let #50 run its course. If it doesn't move forward by the 18th, this PR is ready to pick back up. Appreciate the transparency on the process.

knisar:
@abhnvgrg, thank you for waiting. The September 18 window for #50 has ended without an update, so this PR is now the active candidate for #38. I am removing the ownership block and moving this PR and the issue into review. I will check the current head against current main, including JSON-safe unassessed results and the core pipeline behavior, and leave any concrete findings here.

knisar (changes requested):
Reviewed `a7f226d` integrated with main `592189f`. The normalizer now rejects non-finite values, and the core pipeline correctly excludes those judge results from assessed scores. The existing suite passes: 403 passed, 15 optional tests skipped. Syntax, licensing, and patch whitespace checks pass.

Two changes are needed before merge:

1. **Keep the unassessed result JSON-safe at both judge boundaries.** Both new branches store a float NaN or infinity in `details["raw_score"]`. Even a mocked response of `{"score": "NaN"}` makes `json.dumps(asdict(result), allow_nan=False)` raise `ValueError`. In the shared judge branch, `details["judge_response"]` also retains the original payload, so converting only `raw_score` is insufficient when the incoming score itself is a non-finite float. Preserve useful evidence using strings or another JSON-safe representation throughout the retained payload. Cover NaN and both infinities, supplied as strings and floats, for the shared judge path and `GroundednessScorer`.

2. **Classify these coverage gaps as parser/integration failures.** `_classify_unassessed_reason()` currently returns `"see explanation in JSON report"` for both paths. Set a consistent `skipped="non_finite_judge_score"` marker in the shared judge path as well as groundedness, and register that marker in `rai_toolkit/assessment/assessor.py` to return `"scorer integration / parser failure"`. Add a focused regression for the emitted results, rather than relying on explanation wording.

Minimal offline reproduction for both findings:

```python
import json
from dataclasses import asdict
from unittest.mock import Mock
from rai_toolkit.scorers.llm_judges import FairnessJudge, GroundednessScorer
from rai_toolkit.assessment.assessor import _classify_unassessed_reason

for cls in (FairnessJudge, GroundednessScorer):
    scorer = cls(api_key="offline-test")
    scorer._call_judge = Mock(return_value={"score": "NaN"})
    result = scorer.score("Answer", context="Reference")
    print(_classify_unassessed_reason(result))
    json.dumps(asdict(result), allow_nan=False)
```

Please also include an offline core-pipeline regression for the issue's acceptance criterion: a mocked NaN response is unassessed, cannot supply an assessed pass, and does not dilute a valid result on a second row. I verified that behavior currently works for both paths across all six string/float non-finite inputs; the PR's new tests only exercise the normalizer/result objects and aggregation directly. Preserve finite clamping and inversion as controls.

My 40 focused checks produced 16 passes for pipeline behavior and finite controls, with 24 failures covering the two gaps above. No provider calls were made. Thank you for waiting through the ownership window; these are now the concrete remaining review items.

abhnvgrg:
Thanks for the detailed review @knisar — pushed all three fixes.

JSON safety: Non-finite raw_score values are now stored as string representations. judge_response is sanitized recursively via the existing _sanitize_judge_value helper (promoted to a shared base-class method) — this also covers the assessed path, since a finite score doesn't guarantee sibling fields are finite. json.dumps(asdict(result), allow_nan=False) passes for all six non-finite inputs on both paths.

Classification: Both LLMJudgeScorer.score() and GroundednessScorer.score() now emit skipped="non_finite_judge_score", registered in _classify_unassessed_reason as "scorer integration / parser failure".

Tests: 68 new offline tests covering FairnessJudge and GroundednessScorer × all six non-finite forms (string and float variants of NaN, ±Infinity), plus pipeline regression tests verifying a NaN row doesn't dilute a valid row's score and can't produce an assessed pass. Finite clamping, inversion, and two-row aggregation controls are included. The underlying min(1.0, nan) → 1.0 trap is pinned as a documented test.

All 435 tests pass (14 pre-existing skips). Only ruff findings in touched files are pre-existing BLE001 warnings.

tests/test_nonfinite_scores.py
```diff
@@ -0,0 +1,358 @@
...
+    """Inverted risk scales (higher raw = worse) keep their finite behavior."""
+    assert ScoreNormalizer.from_scale(
+        raw_score, max_value=3.0, invert=True
+    ) == pytest.approx(expected)
+
+
+def test_normalizer_clamp_swallows_nan_which_is_why_scorers_must_reject_it() -> None:
+    """Pins the underlying trap the scorer guard exists to close."""
+    assert ScoreNormalizer.from_compliance_scale(float("nan")) == 1.0
+    assert math.isfinite(ScoreNormalizer.from_compliance_scale(float("-inf")))
```
knisar: [P1] Assert rejection at the shared normalization boundary. This test currently requires `from_compliance_scale(NaN)` to return a perfect score and accepts clamped negative infinity. That preserves the exact bug #38 asks this PR to fix. Restore the `from_scale` finite-input checks and change these assertions to expect `ValueError` for NaN and both infinities, covering non-finite `max_value` as well. Retain the separate judge tests that turn invalid scores into unassessed results.

knisar (changes requested):
Thanks @abhnvgrg. The revision adds the JSON-safe judge payloads, consistent skip marker, and coverage classification requested in the last review. The hosted checks are green.

One blocking regression remains in the revised branch: the shared normalizer guards from the previous implementation are gone. `rai_toolkit/scorers/normalizer.py` is unchanged from main, so the shared API still accepts non-finite scores and scales. I reproduced these results on `893f6e3`:

```python
ScoreNormalizer.from_scale(float("nan"), 3)  # 1.0
ScoreNormalizer.from_scale(float("inf"), 3)  # 1.0
ScoreNormalizer.from_scale(float("-inf"), 3) # 0.0
ScoreNormalizer.from_scale(1, float("nan"))  # 1.0
ScoreNormalizer.from_scale(1, float("inf"))  # 0.0
```

Issue #38 explicitly requires validation at the shared normalization boundary. The judge guards are needed too, but callers of `ScoreNormalizer` must also be protected. Please restore finite validation for both `raw_score` and `max_value`, before clamping and alongside the positive-scale check. Replace the new test that asserts NaN becomes 1.0 with rejection tests for NaN and both infinities in each argument. Keep finite clamping/inversion and the new judge/pipeline regressions.

The inline comment identifies the test currently locking in the bug. Please also refresh the PR description and validation counts to describe the final diff. I checked the diff and targeted offline reproductions; this is not a fresh full-suite validation.

abhnvgrg:
Thanks @knisar
Good catch, the normalizer guards were dropped during a branch reset. Restored math.isfinite validation for both raw_score and max_value in from_scale, and replaced the NaN-as-1.0 pin test with rejection tests for all six non-finite inputs across from_scale and from_compliance_scale. Judge and pipeline regressions are unchanged.

knisar (approved):
Reviewed `a64bfb0658861c655ef3a81fc187fc7994d632d3` integrated with current main `e70df99`. The requested normalizer guards are restored, the incorrect NaN-as-perfect-score test is replaced with rejection tests, and the JSON-safe unassessed results and coverage classification are retained. No blocking findings remain.

Validation: 479 passed and 15 optional tests skipped in the full core suite; 24 additional offline pipeline checks passed for all-invalid and valid-first/invalid-last runs across both judge paths and all six non-finite forms. Lint, REUSE licensing, and patch whitespace checks passed. All nine hosted checks on this head passed.

I refreshed the PR description to match the final implementation and verified results. Approved. Thank you for the revisions and for waiting through the ownership handoff.

PR #72, opened 2026-09-15
feat: add deterministic tool-call accuracy scorer

### LINKED ISSUE #71: Add a deterministic tool-call accuracy scorer for agentic runs
## Problem

First slice of the agentic work tracked by #4. The toolkit has retrieval scorers, but no scorer for whether an agent requested the expected tools with the expected arguments.

`rai_toolkit/evaluation/agentic.py` provides scaffolding, including `TurnSpec` and `AgenticEvalSpec.success_criteria`. It does not define a structured call record or collect execution traces. A `role="tool"` turn is usually a tool result, and call-shaped text in an assistant response can be a quotation or refusal. Neither is sufficient evidence that a call was requested.

## Scope and public interface

Add an offline, deterministic `ToolCallAccuracyScorer` derived from `BaseScorer`, exported from `rai_toolkit.scorers`, with `name="tool_call_accuracy"` and default `threshold=1.0`.

Keep the existing `score(output, input="", context="", **kwargs)` interface and accept these additional keyword arguments:

- `tool_calls`: a list of structured invocation records, or `None` when the trace is unavailable.
- `trace_complete`: an explicit Boolean indicating that the caller captured the complete invocation history for this evaluation window. Only `True` permits assessment.
- `success_criteria`: the dictionary used by `AgenticEvalSpec`, with the `tool_calls` block defined below.

The caller supplies calls from recorded invocation events. Do not extract calls from `output`, `input`, `context`, refusal text, or tool-result prose. This measures requested calls and their arguments, not successful tool execution or task completion.

This first implementation supports direct calls, inherited `score_async`, and `score_batch`. The current evaluation pipeline forwards only selected scorer inputs, so do not claim automatic trace forwarding or integration with `Assessor`, the CLI, or Weave in this issue.

## Input contract

Example:

```python
result = ToolCallAccuracyScorer().score(
    output="The requested record was found.",
    tool_calls=[
        {"id": "call-1", "name": "lookup_record", "arguments": {"record_id": 42}}
    ],
    trace_complete=True,
    success_criteria={
        "tool_calls": {
            "expected": [
                {"name": "lookup_record", "arguments": {"record_id": 42}}
            ],
            "forbidden": ["send_email"],
        }
    },
)
```

- Each actual record has exactly `id`, `name`, and `arguments`. IDs and names are non-empty strings; IDs are unique within the trace. `arguments` is a JSON-compatible dictionary with string keys. Repeated calls with distinct IDs remain distinct observations.
- `success_criteria["tool_calls"]` has a required `expected` list and an optional `forbidden` list, defaulting to `[]`. Reject unknown keys within this block and within call records. Other top-level success criteria can coexist and are ignored by this scorer.
- Each expected record has exactly `name` and `arguments`, with the same name/argument validation. Expected duplicates are allowed and require separate actual calls. Expected records have no runtime call ID.
- `forbidden` contains unique, non-empty, case-sensitive tool names. A name cannot be both expected and forbidden; conflicting criteria are invalid.
- An explicitly present `expected: []` means that no calls are expected. It is a real criterion, distinct from an absent `tool_calls` block or absent `expected` key.
- Accept JSON values only: dictionaries with string keys, lists, strings, finite numbers, Booleans, and null. Reject non-finite numbers, non-JSON objects, cyclic structures, and serialized argument strings. Do not silently coerce malformed records or discard them before scoring.

## Matching and score

1. Match expected calls to actual calls one-to-one using exact tool name and argument equality. Process expectations in their supplied order and choose the earliest unused matching actual call. This fixes evidence selection without requiring execution order to match expectation order.
2. Argument object-key order is irrelevant; key sets must match exactly, including nested objects. Array order matters. Strings are compared exactly, with no trimming or case folding. JSON numbers compare numerically, so `1` equals `1.0`; Booleans remain distinct from numbers, so `True` does not equal `1`. No string-to-number conversion or other coercion is allowed. Missing or extra argument keys produce a mismatch.
3. Let `E` be the number of expected calls, `A` the number of actual calls, and `M` the number of exact matches. The score is the call-matching F1 value `2 * M / (E + A)`. If `E == A == 0`, the score is `1.0` because the explicit no-call criterion was met. Missing, wrong, extra, and surplus duplicate calls therefore lower the score.
4. If any forbidden tool was requested, set `score=0.0` and `passed=False`, regardless of threshold. Otherwise, `passed = score >= threshold`. Reject invalid configured thresholds at construction with `ValueError`: thresholds must be finite numbers in `[0, 1]`, excluding Booleans. The default threshold of `1.0` requires a complete exact match with no extra calls; a caller can explicitly select a lower acceptance threshold for partial matches.

Examples at the default threshold:

| Expectation and complete trace | Score | Result |
| --- | --- | --- |
| One expected call, one exact actual call | 1 | Pass |
| One expected call, no actual calls | 0 | Fail |
| One expected call, one actual call with wrong arguments | 0 | Fail |
| One expected call, two identical actual calls with distinct IDs | 2/3 | Fail |
| Explicitly no expected calls, no actual calls | 1 | Pass |
| Any forbidden call observed | 0 | Fail |

## Assessed versus unassessed

For a complete, valid trace with valid criteria, return `assessed=True`, including when the call list is empty. At the default threshold, an expected call that was never made is an assessed failure, not a skipped row.

Missing, incomplete, or malformed trace data and missing or malformed criteria return `assessed=False`, `score=0.0`, and `passed=False`. Use stable reasons in `details["skipped"]`, such as `missing_tool_trace`, `incomplete_tool_trace`, `invalid_tool_trace`, `missing_tool_call_criteria`, and `invalid_tool_call_criteria`. An absent `expected` key is missing criteria; an explicitly empty list is the no-call scenario above. An omitted completeness flag must not be treated as `True`.

Never return an assessed pass from a partial trace or after dropping invalid records. Direct consumers must check `assessed` before treating the placeholder score as a measurement. Follow the existing [scorer-authoring guide](https://github.com/wandb/rai-toolkit/blob/main/docs/scorer_authoring.md).

## Evidence and diagnostics

Include `scorer_name`, the score's counts (`expected_count`, `actual_count`, `matched_count`), matched expected indexes and actual call IDs/indexes, and diagnostics distinguishing missing calls, argument mismatches, unexpected calls, and forbidden calls.

For diagnostics, first remove exact matches. Pair remaining expectations with remaining actual calls of the same name in input order to identify argument mismatches. Unpaired expectations are missing calls. Unpaired actual calls are unexpected calls. List forbidden calls independently. Diagnostic pairing must not change `M` or count any call twice in the formula.

Evidence must refer to the original supplied records and expected indexes, with original argument values preserved. Do not invent evidence for a missing call. Keep all result details JSON-safe with strict serialization (`allow_nan=False`), including when the rejected input contains non-finite values or non-JSON objects. Report the validation location and reason without copying invalid raw values into the result. Do not mutate caller-owned inputs.

## Acceptance tests

Add offline tests for:

- Exact matches; reordered object keys; reordered independent calls; nested dictionaries and ordered arrays; numeric, Boolean, and string distinctions.
- Missing calls, missing/extra/wrong argument keys, unexpected calls, and forbidden calls. Forbidden calls must fail even with a threshold of zero.
- Repeated expectations and repeated actual calls, one-to-one matching, deterministic diagnostic pairing, and unique-ID validation. Surplus duplicate calls must not improve the score.
- Complete empty traces with non-empty expectations versus an explicit empty expectation list.
- Missing, incomplete, and malformed traces and criteria; contradictory expected/forbidden names; missing completeness flags; invalid thresholds; non-finite and non-JSON inputs.
- Refusal or quoted call syntax in response text contributing zero observed calls. A complete empty trace still fails when a call was expected, regardless of the response's wording.
- Exact result counts, `assessed`/`passed`, original-record evidence, strict JSON serialization, and preservation of input objects.
- Equivalent behavior for direct `score`, inherited `score_async`, and `score_batch` calls; public import from `rai_toolkit.scorers`.

## Non-goals

No LLM judge, prompt changes, runtime dependency, tool execution, trace collection, provider-specific parser, fuzzy or subset argument matching, trajectory/order grading, task-completion scoring, automatic compliance mapping, or pipeline/Assessor/Weave integration. This can proceed independently of #27 and #62.

## Likely files and documentation

- `rai_toolkit/scorers/tool_calls.py` (new scorer and private validation/matching helpers)
- `rai_toolkit/scorers/__init__.py`
- `tests/test_tool_call_accuracy.py`
- A short README entry and executable direct-use example describing an opt-in programmatic scorer and its integration limits

Do not redesign or re-export the existing agent protocols solely for this scorer.

## Local validation and starting work

Follow `CONTRIBUTING.md` for the lightweight setup and required checks. Run the focused tests, full core suite, lint, licensing, and whitespace checks; include exact commands and results in the PR. No API keys or network calls are needed by these tests.

Before starting, check this issue's Development section and the open PRs for linked work. Once you have an initial commit, open a draft PR with `Closes #NN` using this issue's number, the implementation approach, and your test plan. The linked draft signals that work has started; no `/claim` comment or status-label wait is needed.

adity982:
Closes #71

## Summary
- add an offline `ToolCallAccuracyScorer` for complete structured agent invocation traces
- validate JSON-safe records and criteria without coercion, including precise safe validation paths and reasons
- match expected calls one-to-one with deterministic F1 scoring, forbidden-call handling, and original argument evidence
- distinguish missing expected criteria from malformed criteria and handle arbitrarily large JSON integers without float conversion
- export the scorer, document direct opt-in use and current pipeline limitations, and add focused offline regression coverage

## Validation
- `py -3 -X utf8 -m pytest -q tests/test_tool_call_accuracy.py` — 18 passed (14 warnings; existing Python 3.14 deprecations/config warning)
- `py -3 -X utf8 -m ruff check rai_toolkit/scorers/tool_calls.py tests/test_tool_call_accuracy.py` — passed with locally available Ruff 0.16.7
- `git diff --check origin/main...HEAD` — passed

A local full-suite attempt without installing new dependencies reported 336 passed and 10 skipped; 70 async tests could not run because `pytest-asyncio` is absent, and one package-metadata test could not run because this checkout is not installed as a distribution. REUSE is also unavailable locally. The PR's hosted core, quality/licensing, optional Weave, and package checks remain the authoritative full-environment validation.

This is an AI-assisted contribution; no network calls or API keys are required by the new tests.


rai_toolkit/scorers/tool_calls.py
```diff
@@ -0,0 +1,180 @@
...
+                remaining_actual.remove(actual_index)
+                remaining_expected.remove(expected_index)
+                mismatches.append({"expected_index": expected_index, "actual_index": actual_index})
+        details = {
+            "scorer_name": self.name,
+            "expected_count": len(parsed_expected),
+            "actual_count": len(actual),
+            "matched_count": len(matched_expected),
+            "matched_expected_indexes": matched_expected,
+            "matched_actual_ids": matched_ids,
```
knisar: **[P2] Include original argument values in the evidence.** `details` currently keeps counts, IDs, and indexes but discards the argument values required by #71. Two exact matches with the same call ID but `arguments={"record_id": "record-A"}` versus `{"record_id": "record-B"}` produce identical result details; an argument mismatch likewise contains only two indexes. A saved result therefore cannot show what was compared without the original caller objects. Include JSON-safe evidence tied to the original expected indexes and actual IDs/indexes, preserving supplied argument values for matched calls and relevant diagnostics. Do not invent actual-call evidence for a missing expectation, and keep caller-owned inputs unchanged.

rai_toolkit/scorers/tool_calls.py
```diff
@@ -0,0 +1,180 @@
...
+
+from rai_toolkit.scorers.base import BaseScorer, ScorerResult
+
+
+def _json_value(value: Any, seen: set[int] | None = None) -> bool:
+    """Return whether *value* is composed only of strict JSON values."""
+    if value is None or isinstance(value, (str, bool)):
+        return True
+    if isinstance(value, (int, float)):
+        return not isinstance(value, bool) and math.isfinite(value)
```
knisar: **[P2] Validate integers without converting them to floating point.** `math.isfinite(10 ** 400)` raises `OverflowError`, although this is a finite integer accepted by `json.dumps(..., allow_nan=False)`. Supplying that value in matching actual/expected arguments currently raises from `score()` instead of returning the required assessed match. The same conversion in threshold validation makes `ToolCallAccuracyScorer(threshold=10 ** 400)` raise `OverflowError` rather than the specified `ValueError`. Handle integer finiteness separately from float finiteness, check threshold bounds without an overflowing conversion, and cover both paths.

rai_toolkit/scorers/tool_calls.py
```diff
@@ -0,0 +1,180 @@
...
+        input: str = "",
+        context: str = "",
+        **kwargs: Any,
+    ) -> ScorerResult:
+        tool_calls = kwargs.get("tool_calls")
+        if tool_calls is None:
+            return _unassessed(self.category, self.name, "missing_tool_trace", "Tool invocation trace is unavailable.")
+        if kwargs.get("trace_complete") is not True:
+            return _unassessed(self.category, self.name, "incomplete_tool_trace", "Tool invocation trace was not marked complete.")
+        if not isinstance(tool_calls, list) or not _json_value(tool_calls):
```
knisar: **[P2] Return the failing validation location and reason.** The Boolean-only JSON validator loses the position of rejected data. For a non-finite value nested in `tool_calls[0].arguments.value.nested_key`, the result only says the trace is malformed; the expected-criteria path has the same problem. #71 requires the validation location and reason without copying invalid raw values into result details. Propagate a safe path plus a specific reason such as `non_finite_number`, including the affected record index and argument path, and retain strict JSON serialization for these rejected inputs.

knisar (commented):
Early feedback on draft `6a2a310` against #71. The structured-trace approach, strict completeness flag, one-to-one matching, Boolean/number distinction, F1 formula, forbidden-call override, and deterministic mismatch pairing look consistent with the requested contract in the cases I exercised.

All existing CI checks are green. Locally, the core suite passed with **331 passed, 15 skipped** on the PR head and **342 passed, 15 skipped** after integrating current main `9f7091f`. A separate set of **46 contract checks had 38 passes and 8 failures**. The three inline comments cover the main gaps: original argument evidence, integer validation that can raise instead of returning a result, and validation locations.

Two additional completion items before marking this ready:

- With a complete empty trace and `success_criteria={"tool_calls": {}}`, `details["skipped"]` is currently `invalid_tool_call_criteria`. #71 explicitly treats an absent `expected` key as missing criteria, so this should be `missing_tool_call_criteria`. Preserve the distinction from an explicit `expected: []`.
- Add the executable direct-use example and integration limitations requested in #71. This is an opt-in programmatic scorer; the current pipeline does not forward invocation traces automatically.

Please expand the checked-in tests to cover these failures. The 38 passing review cases also exercised duplicate expectations and calls, deterministic pairing, numeric versus Boolean values, nested arguments, ordered arrays, cycles and other invalid JSON values, empty versus incomplete traces, lower thresholds, forbidden calls, input preservation, and async/batch equivalence. Keeping representative cases from these groups will protect the contract as the implementation evolves. Leaving this as draft feedback while you finish the implementation.

rai_toolkit/scorers/tool_calls.py
adity982: Addressed in da1edb0. Matched, mismatched, unexpected, and forbidden calls now retain JSON-safe original argument evidence with expected indexes plus actual IDs/indexes. Missing expectations include expected evidence only, and regression coverage verifies the caller-owned inputs remain unchanged.

rai_toolkit/scorers/tool_calls.py
adity982: Addressed in da1edb0. Integer validation no longer calls math.isfinite, float finiteness remains explicit, and threshold bounds avoid float conversion. Regressions cover a matching 10**400 argument and a 10**400 threshold returning ValueError rather than OverflowError.

rai_toolkit/scorers/tool_calls.py
adity982: Addressed in da1edb0. Strict JSON validation now returns a safe location and reason (for example tool_calls[0].arguments.value.nested_key / non_finite_number) without copying the rejected raw value. Tests cover both actual-trace and expected-criteria paths and assert strict serialization of the result details.

adity982:
Addressed the remaining draft feedback in `da1edb0`:

- an absent `expected` key now returns `missing_tool_call_criteria`, while explicit `expected: []` remains valid;
- README now includes an executable direct-use example and states that the current pipeline does not forward invocation traces automatically;
- checked-in regressions now cover original argument evidence, input immutability, large integers, safe validation locations/reasons, deterministic duplicate pairing, and missing-vs-invalid criteria.

Focused validation: 18 scorer tests passed, touched-file Ruff passed, and the branch diff check passed. The local full-suite attempt was environment-limited as recorded in the PR body; hosted core and quality/licensing checks are green on this exact head, with the remaining hosted jobs still running at the time of this update.

knisar (commented):
Rechecked `da1edb0` and integrated it with current main `f4328ef` after #73 merged. The previous findings are resolved: original argument evidence and call indexes are retained, large integers avoid float conversion, invalid input has safe validation locations/reasons, absent expected criteria use the missing-criteria reason, and the README provides a working direct-use example with accurate integration limits.

Validation on the integrated tree: **432 passed, 15 skipped**, including all **46 independent contract checks**. The README example, Ruff, REUSE (137/137 files), and working-tree/staged/PR whitespace checks passed. All seven hosted CI jobs and both security checks are green on `da1edb0`.

I have no remaining correctness findings from this review. @adity982, please mark the PR ready for review when you consider it complete. Once it leaves draft, I will perform the final head/CI check, approve, and merge.

knisar (approved):
Final check: the author has marked this PR ready, and `da1edb0` is unchanged from the follow-up review. Main remains `f4328ef`, the exact base used for the successful integration check.

The integrated suite passed with 432 tests and 15 skips, including all 46 independent contract checks. The README example, Ruff, REUSE, and patch-whitespace checks passed. All seven hosted CI jobs and both security checks remain green. The previous review findings are resolved. Approving for merge.

PR #74, opened 2026-09-16
fix(scorers): report an unsupported expected_format as unassessed

KunyangZhang:
## What changed

`OutputFormatScorer` reports an unsupported `expected_format` as an unassessed
row instead of a full pass. Adds offline tests for the scorer.

## Why

`expected_format` is compared case-sensitively against `json` and `xml`, and
every other value took this branch:

```python
else:
    return ScorerResult(
        score=1.0,
        passed=True,
        category=self.category,
        explanation=f"Unknown format '{self.expected_format}', skipping check",
    )
```

So a configuration mistake - `expected_format="JSON"`, `"yaml"`, or a typo -
scored a full pass for **any** output, including empty and malformed ones, with
`assessed=True`:

```python
>>> OutputFormatScorer(expected_format="JSON").score("not json at all")
ScorerResult(score=1.0, passed=True, ..., assessed=True)
```

A row that was never checked then counts as a measured pass in aggregates, which
is the opposite of what the result contract asks for - `assessed` is "True only
when the scorer produced a real signal ... surface the gap honestly instead of
inflating with neutral defaults". The scorer-authoring guide specifies the
unassessed shape: `score=0.0`, `passed=False`, `assessed=False`, and a stable
`details["skipped"]` reason. The branch now returns that, and records
`scorer_name` in `details` as the guide also asks. The `json` and `xml` paths are
unchanged, so no correctly configured scorer changes behaviour.

## How it was checked

```
python -m pytest -q                                         # 332 passed, 15 skipped
python -m pytest tests/test_output_format_scorer.py -q       # 11 passed
python -m ruff check --select E9,F63,F7,F82 .                # All checks passed!
git diff --check                                            # clean
```

The new tests cover valid JSON, invalid JSON, a missing required key, valid and
invalid XML, and the unsupported-format row including its coverage-gap effect.
Six of them fail on the current `main`.

Notes:

- `reuse --no-multiprocessing lint` was not run: this environment is offline and
  `reuse==6.2.0` is not installable from it. The new test file carries the SPDX
  header in the same form as the other test files, with my own name as the
  copyright holder. The `ruff` available here is 0.16.7 rather than the 0.16.6
  pinned in CI, on the `E9,F63,F7,F82` selection.
- This is independent of #73, which fixes false keyword hits in
  `KeywordToxicityScorer`; both touch `programmatic.py`, in different places. If
  you would rather review a single change for the module, I can fold them
  together.

This contribution was prepared with AI assistance, as the contributing guide
asks contributors to disclose.


knisar (approved):
Reviewed `21ee7ea` against main `5439934`. The unsupported-format branch now returns the documented unassessed result with a stable reason, and the JSON and XML paths are unchanged.

Validation: the core suite passed with **332 passed, 15 skipped**. Another **11 review checks passed**, covering unsupported formats at thresholds 0, 0.5, and 1 across empty, malformed, and valid JSON output, plus the actual evaluation pipeline. In the pipeline, an unsupported format is counted as a coverage gap and excluded from the measured average; a run containing only unsupported formats has no measured pass rate. All seven CI jobs and both security checks passed.

No blocking findings. Thanks @KunyangZhang for the focused fix and regression coverage.

PR #83, opened 2026-09-18
feat(cli): report the toolkit version with --version

NourMoftah:
Closes #80

## Summary
- Add a top-level `rai --version` argparse action that prints `rai <version>` from `rai_toolkit._version.__version__` and exits without a subcommand.
- Do not read `importlib.metadata.version()` for this path, so a source checkout still reports the same authority as `rai_toolkit.__version__`.

## Test plan
- [x] `python -m pytest -q tests/test_cli_version.py tests/test_cli.py tests/test_version.py` (13 passed)
- [x] `python -m pytest -q` (389 passed, 15 skipped with `.[dev]` only)
- [x] `python -m ruff check --select E9,F63,F7,F82 .`
- [x] `reuse --no-multiprocessing lint`
- [x] `git diff --check origin/main...HEAD`

Prepared with AI assistance.

PR #84, opened 2026-09-18
fix(cli): fail policy lint when no YAML files were checked

### LINKED ISSUE #81: fix(cli): fail policy lint when no YAML files were checked
## Problem

On main `592189f`, `rai policies lint <directory>` exits successfully and prints `OK: 0 policies loaded` when the directory contains no matching policy files. A CI check can therefore pass without validating any policies.

The existing CLI tests from #41/#47 cover valid and invalid policy files, but not an empty directory or a directory containing only unrelated files.

## Reproduction

No API keys or external services are needed:

```python
from pathlib import Path
from tempfile import TemporaryDirectory
from rai_toolkit.cli import main

with TemporaryDirectory() as directory:
    print("exit:", main(["policies", "lint", directory]))
    Path(directory, "README.txt").write_text("No policy files here.")
    print("exit:", main(["policies", "lint", directory]))
```

Both calls currently return 0. Both should return 1 because the CLI checked no policy files.

## Scope and acceptance criteria

Make a narrow change to `_cmd_policies_lint` in `rai_toolkit/cli.py`:

- If loading succeeds but no policy sets were loaded, print a concise `FAIL:` diagnostic to stderr that includes the requested directory and explains that no `*.yaml` policy files were found. Return 1 and do not print an `OK:` summary.
- Preserve successful linting when at least one valid policy set was loaded, even if it declares `policies: []` or all its policies have `enabled: false`. Those files were actually validated.
- Preserve success output and policy listings for normal valid files, and preserve existing failure behavior for malformed YAML/schema data, missing directories, and file paths passed as directories.
- Keep the current nonrecursive `*.yaml` discovery behavior. This task does not add `.yml` support or policy-pack manifest support.
- Keep `PolicyEngine.from_directory()` unchanged and usable for callers that intentionally allow an empty engine. The stricter requirement belongs to the lint command.

Implementation hint: the engine exposes `policy_sets`, which records loaded files. `engine.policies` contains only enabled policies, so checking that list would incorrectly reject valid disabled-only sets.

## Tests and likely files

- Implementation: only the lint command handler in `rai_toolkit/cli.py`.
- Add `tests/test_cli_policy_lint_empty.py` with offline `tmp_path` fixtures and calls to `main()`.
- Cover an empty directory, a directory with only unrelated files, a valid empty set, a valid disabled-only set, and a normal enabled policy.
- For the new failures, assert exit code 1, the stderr reason/path, and absence of `OK:` on stdout. For valid files, assert exit code 0 and the existing success behavior.
- Run the existing invalid-policy test as a regression control.

Run `python -m pytest -q tests/test_cli_policy_lint_empty.py tests/test_cli.py`, then the checks in `CONTRIBUTING.md`.

## Non-goals and starting work

No policy evaluation/schema changes, policy-pack redesign, new runtime dependencies, or changes to other CLI handlers. This is independent of the enterprise policy-pack design tracked in #3.

This is a scoped starter task requiring basic Python conditionals and pytest fixtures. Check for linked work first, then open a draft PR early with `Closes #NN` and your test plan. Use this issue's number and follow `CONTRIBUTING.md`; the linked draft signals that work has started.

NourMoftah:
Closes #81

## Summary
- Fail `rai policies lint` when loading succeeds but no policy sets were found, so an empty directory or a directory with only unrelated files cannot pass CI.
- Print a `FAIL:` diagnostic on stderr that includes the requested path and that no `*.yaml` policy files were found. Do not print an `OK:` summary in that case.
- Keep success for valid files, including empty policy lists and disabled-only sets. `PolicyEngine.from_directory()` is unchanged.

## Test plan
- [x] `python -m pytest -q tests/test_cli_policy_lint_empty.py tests/test_cli.py` (12 passed)
- [x] `python -m pytest -q` (392 passed, 15 skipped with `.[dev]` only)
- [x] `python -m ruff check --select E9,F63,F7,F82 .`
- [x] `reuse --no-multiprocessing lint`
- [x] `git diff --check origin/main...HEAD`

Made with [Cursor](https://cursor.com)


knisar (approved):
Reviewed 19d2aca. The lint command rejects directories with no loaded YAML policy sets and preserves valid empty and disabled-only sets. The engine API is unchanged. Local validation: 392 passed, 15 optional tests skipped; syntax checks, license checks, and patch whitespace passed. All nine GitHub checks passed. Approved.

Also tested the proposed merge with main ce2995a after #83 landed: 395 passed, 15 optional tests skipped, with clean patch whitespace.

PR #85, opened 2026-09-18
fix(cli): fail policy lint when no YAML files were checked

HarshRajSinghania:
Closes #81

## Summary
Make `rai policies lint <directory>` fail when loading succeeds but no policy sets were discovered. An empty directory or a directory with only unrelated files previously printed `OK: 0 policies loaded` and exited 0, which could let a CI lint step pass without checking any policies.

## Motivation
Maintainer-scoped request in #81 (`status: available`). `PolicyEngine.from_directory()` is left unchanged for callers that allow an empty engine. The stricter check belongs only to the lint command.

## Implementation
In `_cmd_policies_lint`, after a successful load, if `engine.policy_sets` is empty print a `FAIL:` diagnostic on stderr that includes the requested directory and that no `*.yaml` files were found, then return 1. Do not print an `OK:` summary in that case.

Valid files that load as an empty set (`policies: []`) or as disabled-only policies still succeed, because those files were actually validated. Discovery remains nonrecursive `*.yaml` only.

## Testing
Added `tests/test_cli_policy_lint_empty.py` covering:
- empty directory (exit 1, stderr FAIL + path, no OK on stdout)
- directory with only unrelated files
- valid empty policy set (exit 0)
- valid disabled-only set (exit 0)
- normal enabled policy (exit 0)
- existing invalid-policy regression (exit 1)

Commands run:
```
PYTHONPATH=. python3 -m pytest -q tests/test_cli_policy_lint_empty.py tests/test_cli.py
```
Result: 12 passed.

Full `.[dev]` install and `reuse lint` were not run in this environment because the sandbox PyPI mirror failed during hatchling install. The change is limited to the lint handler and a dedicated test file.

PR #89, opened 2026-09-19
fix(cli): fail policy lint when no YAML files were checked

### LINKED ISSUE #81: fix(cli): fail policy lint when no YAML files were checked

HarshRajSinghania:
## Summary

`rai policies lint <directory>` now exits 1 when the directory contains no `*.yaml` policy files, instead of printing `OK: 0 policies loaded` and exiting 0.

## Motivation

Closes #81. A CI job that pointed the lint command at an empty or unrelated directory could pass without validating any policies.

## Implementation

In `_cmd_policies_lint`, after a successful `PolicyEngine.from_directory()` call, treat an empty `engine.policy_sets` list as a lint failure. The diagnostic is written to stderr as `FAIL: no *.yaml policy files found in <directory>` and no `OK:` summary is printed.

`PolicyEngine.from_directory()` is unchanged, so callers that intentionally allow an empty engine still can. Valid files that declare `policies: []` or only `enabled: false` policies still succeed, because those files were loaded into `policy_sets`.

Discovery remains non-recursive `*.yaml` only.

## Testing

Added `tests/test_cli_policy_lint_empty.py` covering:

- empty directory
- directory with only unrelated files
- valid empty policy set
- valid disabled-only policy set
- normal enabled policy
- malformed policy (regression)

Commands run:

```
python -m pytest -q tests/test_cli_policy_lint_empty.py tests/test_cli.py
```

Result: `12 passed`.

Full `.[dev]` install was not used in this environment because the configured package index returned 502s for build backends. Tests were executed with `PYTHONPATH` after installing `openai`, `pyyaml`, and `pydantic` from PyPI.


knisar:
@HarshRajSinghania, this overlaps with your #85 and the earlier #84 for #81. I have explained the review order on #85: #84 will be reviewed first. Please pause further work here while that review runs. I will follow up with the outcome so you are not left waiting. Thank you for contributing.

knisar:
Closing this overlapping implementation of #81 because #84 has now passed review and CI and merged. Thank you, @HarshRajSinghania, for the contribution. The review-order explanation is on #85.

